PR #31, opened 2013-11-24
DATAES-33: Mapping of parent-child relationships

phjardas:
Elasticsearch allows parent-child relationships for documents, see http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/mapping-parent-field.html
- The parent document requires no special treatment.
- The child document has to be configured with a special mapping indicating the type of the parent document.
- Each inserted child document must have a reference to the ID of its parent document.

Here's my approach:
- We need a new field annotation (my suggestion: @ParentId) that marks the field which contains a child document's parent ID. Only one field may be annotated with this.
- The mapping builder needs to be updated to handle this new annotation.
- ElasticsearchTemplate queries need to be able to insert child documents with a parent references.

see https://jira.springsource.org/browse/DATAES-33

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Parent.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Parent.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


ft0907:
There was a problem when using @parent annotations：
`nested exception is java.lang.IllegalArgumentException: can't add a _parent field that points to an already existing type, that isn't already a parent`
@Getter
@Setter
@Document(indexName = "#{indexName}", type = "content")
public class Content extends BaseEntity {
        @NotNull
	@Parent(type = "contentcategory")
	@Field(type = FieldType.Keyword)
	private String contentCategoryId;
}

@Getter
@Setter
@Document(indexName = "#{indexName}", type = "contentcategory")
public class ContentCategory extends BaseEntity{
	@NotEmpty
	@Length(max = 200)
	@Field(type = FieldType.Text, searchAnalyzer = "ik_max_word", analyzer = "ik_max_word")
	private String title;
}
I can't know what the above error is caused by.
Can you give me a answer? 
Thanks

PR #63, opened 2014-07-02
DATAES-100: Allow configurable searchTimeout.

greyfairer:
In ES Client, in the actionGet(), you can specify a timeout, after which you will get a ElasticSearchTimeoutException.
If you use the actionGet() without timeout, some calls will block indefinitely or until the thread is interrupted.
I'd suggest to make it a single property of ElasticsearchTemplate, to be used in all search queries.


akonczak:
rebased and merge to master

PR #72, opened 2014-08-26
DATAES-115 Add count before projection query to retrieve all results.

kevinleturc:


kevinleturc:
Hi,
Any news on this ?
Regards.

PR #75, opened 2014-09-01
DATAES-127 - Include total, other and missing count in TermResult

mobreza:
Include totalCount, otherCount and missingCount from TermsFacet in TermResult.


akonczak:
rebased and merged

PR #78, opened 2014-09-12
DATAES-129 : Custom Repository Method for simple geo request does not wo...

fmarchand:
If you have a field named "Location" which is a geo-point and if you call a method in a custom repository called "findByLocation" then it will try to execute a query_string instead of doing a "match all" with a geo filter. It means the type of the field is not detected as a geo-point.

I just noticed that there is no query or filter in ES to retrieve an exact geopoint so I used a trick : I used a "geo distance" filter with a distance of 0.001km.


mohsinh:
merged ! thanks :)

PR #82, opened 2014-10-25
DATAES-132 support include_in_parent for nested fieldtype

martinsachs:
i added another annotation attribute for this feature. Defaults to false, but only written if type is 'Nested'


mohsinh:
merged

PR #90, opened 2014-12-31
DATAES-143 Auto Index creation should be configurable

mason:
When the application is loaded, indexes are created automatically from the Entity definitions/annotations. This should be an optional setting as automatically generating indexes interferes with template definitions.

Example:
@Document(indexName = "test-index", type = "test-type", createIndex = "false")

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Document.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Document.java
@@ -24,6 +24,7 @@ import org.springframework.data.annotation.Persistent;
  *
  * @author Rizwan Idrees
  * @author Mohsin Husen
+ * @author Mason Chan
  */
 
 @Persistent
@@ -43,4 +44,8 @@ public @interface Document {
 	String refreshInterval() default "1s";
 
 	String indexStoreType() default "fs";
+
+	String createIndex() default "true";
+
+
 }
--- a/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
@@ -104,6 +104,7 @@ import static org.springframework.data.elasticsearch.core.MappingBuilder.buildMa
  * @author Mohsin Husen
  * @author Artur Konczak
  * @author Kevin Leturc
+ * @author Mason Chan
  */
 
 public class ElasticsearchTemplate implements ElasticsearchOperations, ApplicationContextAware {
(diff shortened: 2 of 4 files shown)


mason:
Hello, any news on this?

GuillermoBlasco:
This feature is useful in scenarios where index creation and mapping is managed out of the box. But i think that the condition check fits better on `AbstractElasticsearchRepository` constructor, instead of in `ElasticsearchTemplate` since in that case any explicit call to create index shall fall in that checking. We want to disable the automatic index creation and/or the mapping creation for the given type. Therefore checking the condition on `AbstractElasticsearchRepository` [(link)](https://github.com/spring-projects/spring-data-elasticsearch/blob/a3828c6161daa9fe31eacdcc93a286119ac470fd/src/main/java/org/springframework/data/elasticsearch/repository/support/AbstractElasticsearchRepository.java#L74) is the right way since is where the automatic index and mapping creation happens:

```java
	public AbstractElasticsearchRepository(ElasticsearchEntityInformation<T, ID> metadata,
										   ElasticsearchOperations elasticsearchOperations) {
		this(elasticsearchOperations);
		Assert.notNull(metadata);
		this.entityInformation = metadata;
		setEntityClass(this.entityInformation.getJavaType());
		try {
			createIndex();
			putMapping();
		} catch (ElasticsearchException exception) {
			LOGGER.error("failed to load elasticsearch nodes : " + exception.getDetailedMessage());
		}
	}
```

mohsinh:
@GuillermoBlasco is right, proposed implementation is not complete one. Having a  createIndex field in the document is ok but it has to be controlled in a way that when the context is loaded no index is created with this field marked with false. 

We will modify your PR and merge it to the master soon.

philnate:
Is there any news on this or a possible work-around to not have auto created the indexes?

mason-uptake:
Hey @philnate our workaround was to initially create set the indexes to create trash indexes on startup and then after it's done, it get switched to the real indexes. I think we are using spring expressions to dynamically change the indexes.

martin-g:
:+1: for this feature

akonczak:
Hi @mason 
Thank you for your contribution
I did rebase and modified your code commit [branch/DATAES-143](https://github.com/spring-projects/spring-data-elasticsearch/tree/DATAES-143).

Changes from my side:
Added check do flag 'createIndex' into AbstractElasticsearchRepository.
Reverted changes in ESTemplate class.
 
Can you sign [contributor agreement](https://support.springsource.com/spring_committer_signup) ? (Please ignore this request if you did it earlier)

mason:
@akonczak Hey I already signed it when I did the PR. was like in 2014 sometime.

PR #88, opened 2014-12-28
DATAES-141 Mapping Context is null in DefaultResultMapper when using custom Entitymapper

kevinleturc:

PR #95, opened 2015-02-03
DATAES-98 Jackson's ObjectMapper is hard to customize

greyfairer:
This allows to use a custom ObjectMapper in the DefaultEntityMapper.

Please note that this won't work without DATAES-141 (Pull Request #88)

E.g:

```
<bean name="elasticsearchTemplate"
     class="org.springframework.data.elasticsearch.core.ElasticsearchTemplate">
    <constructor-arg name="client" ref="client"/>
    <constructor-arg name="entityMapper" ref="entityMapper"/>
</bean>
<bean name="entityMapper" class="org.springframework.data.elasticsearch.core.DefaultEntityMapper">
        <constructor-arg name="objectMapper" ref="customObjectMapper"/>
</bean>
    <bean id="customObjectMapper" class="org.springframework.http.converter.json.Jackson2ObjectMapperFactoryBean">
        <property name="featuresToEnable">[...]</property>
        <property name="featuresToDisable">[...]</property>
</bean>
```

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/DefaultEntityMapper.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/DefaultEntityMapper.java
@@ -15,10 +15,10 @@
  */
 package org.springframework.data.elasticsearch.core;
 
-import java.io.IOException;
-
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.IOException;
 
 /**
  * DocumentMapper using jackson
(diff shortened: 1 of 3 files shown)


benneq:
+1
even if the current ObjectMapper can be easily copy & pasted with its two configurations, this should be possible.

greyfairer:
I see that DATAES-141 was merged recently. Any news on this one?

karayv:
+100500 
I cannot use java 8 time (jsr310). I need to have ability to do this

`
        objectMapper.registerModule(new JavaTimeModule());
`

sothawo:
I don't think that using inheritance is the right way to accomplish this. Especially not the Spring-like one. Please See my [comment in Jira](https://jira.spring.io/browse/DATAES-98?focusedCommentId=183716&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-183716) of how this can be achieved with injection.

PR #96, opened 2015-02-06
close the node when the node-client is destroyed, so it won't stall the ...

gde-vt:
...app shutdown.


mohsinh:
nodeClient.close() does not do anything at all. 

check https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/client/node/NodeClient.java

also by default node client will store data on physical storage unless you specify indexStoreType="memory" at Document level. 

closing PR now, reopen if needed.

PR #100, opened 2015-04-21
Added missing include_in_root property to Field annoation

archie-sh:
Hi, I use spring-data-elasticsearch in my project and require the inlcude_in_root property for one of my mappings so I baked it in into the code. Would be nice to have it in the distribution.


xhaggi:
Closed because `include_in_root` was dropped in elasticsearch.

PR #103, opened 2015-04-24
DATAES-156 - Nested and Multi-field should not re-use field name

gamars:
I would like to use spring-data-elastic search as part of a larger data-project.
While I was pleasantly surprised to see the fine-grained control the @Field annotation is providing I am a bit confused by the repeating field.name repetition as part of the multi-field and nested Types:
in MappingBuilder.java
```
 private static void addNestedFieldMapping(...) throws IOException {
    builder.startObject(field.getName() + "." + annotation.dotSuffix());
    ...
 }
```
should it not read:
```
private static void addNestedFieldMapping(...) throws IOException {
   builder.startObject(annotation.suffix());
   ...
}
```

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/NestedField.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/NestedField.java
@@ -27,7 +27,7 @@ import java.lang.annotation.Target;
 @Target(ElementType.FIELD)
 public @interface NestedField {
 
-	String dotSuffix();
+	String suffix();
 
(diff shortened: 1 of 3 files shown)


fpompermaier:
+1

khangal:
+1

sokersten:
+1

skromberg:
+1

mohsinh:
we have added this changes while upgrading to elasticsearch 2.x. 

Thanks for the PR

PR #116, opened 2015-08-20
Add support for IndicesOptions in search queries.

bitmelt:
This is important for use cases where there's one index per day or one index per month for time-based data.  Let's say we have a daily index, like `logs-20150820`.  When we want to search a date range, we have to specify multiple indices to search.  To search logs from August 18 to August 20, we would need to search indices `logs-20150818`, `logs-20150819` and `logs-20150820`.  This works fine using `ElasticsearchTemplate` using

```
new NativeSearchQueryBuilder().withIndices("logs-20150818", "logs-20150819", "logs-20150820")
```

However, let's say that we had a problem on August 19th and no logs were indexed, so the index `logs-20150819` doesn't exist, but the other indices do.  Trying to do this search using `ElasticsearchTemplate` results in

```
org.elasticsearch.indices.IndexMissingException: [logs-20150819] missing
```

Elasticsearch has support for ignoring missing indices for this use case.  This is done by passing `IndicesOptions.lenientExpandOpen()` to `org.elasticsearch.action.search.SearchRequestBuilder.setIndicesOptions()`

IndicesOptions has other values as well.  This pull request adds support for passing arbitrary IndicesOptions values with the search query.


xhaggi:
Fixed by #196

PR #119, opened 2015-09-23
DATAES-200 Set persistent entity index and type in query when no indi…

kevinleturc:
…ces/types are set.

Please see https://jira.spring.io/browse/DATAES-200


gulimran:
Hi,

It seems this fix is not merged in 1.3.1.RELEASE. Any idea which release this will be part of and when will that release be available? Many thanks.

PR #121, opened 2015-10-07
Typo in error message

nfrancois:
The message is
`Use epxression or mulitple clauses instead.`
instead of :
`Use expression or multiple clauses instead.`


mohsinh:
merged

PR #122, opened 2015-10-19
Improvements for the ElasticsearchRepository

xhaggi:
This PR improves the following:

* `ElasticsearchRepository.index()` now does not implicit call refresh, which makes no sense IMO
* new overloads for `index()` to support bulk indexing without using `ElasticsearchTemplate`
* expose `refresh()` from `ElasticsearchTemplate` over `ElasticsearchRepository`

@olivergierke let me know if we need a Jira Issue for this.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/repository/ElasticsearchRepository.java
+++ b/src/main/java/org/springframework/data/elasticsearch/repository/ElasticsearchRepository.java
@@ -40,6 +40,10 @@ public interface ElasticsearchRepository<T, ID extends Serializable> extends Ela
 
 	<S extends T> S index(S entity);
 
+	void refresh();
+
+	void refresh(boolean waitForOperation);
+
 	Iterable<T> search(QueryBuilder query);
 
 	FacetedPage<T> search(QueryBuilder query, Pageable pageable);
--- a/src/main/java/org/springframework/data/elasticsearch/repository/ElasticsearchRepository.java
+++ b/src/main/java/org/springframework/data/elasticsearch/repository/ElasticsearchRepository.java
@@ -16,6 +16,7 @@
 package org.springframework.data.elasticsearch.repository;
 
 import java.io.Serializable;
+import java.util.List;
 
 import org.elasticsearch.index.query.QueryBuilder;
 import org.springframework.data.domain.Page;
@@ -32,6 +33,10 @@ import org.springframework.data.repository.NoRepositoryBean;
  */
 @NoRepositoryBean
 public interface ElasticsearchRepository<T, ID extends Serializable> extends ElasticsearchCrudRepository<T, ID> {
+
+	<S extends T> List<S> index(List<S> entities);
+
+	<S extends T> Iterable<S> index(Iterable<S> entities);
 
 	<S extends T> S index(S entity);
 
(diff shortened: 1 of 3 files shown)


mohsinh:
Thanks for the PR, yes we definitely need JIRA issue for each and every commit.

xhaggi:
before I'll create the Jira issues could you please let me know if this make sense in general? If you reject the PR we don't need the Jira issues ;)

olivergierke:
It's usually the other way round: it's more likely we reject a PR than it is that we reject an issue as we might agree that there's an issue but don't necessarily agree on the way to implement the issue :) (which exactly is the case for me in the current situation).

Regarding @mohsinh's comment on the JIRA per commit. That's not necessarily true, but points to a problem of the PR: "Improvements to ElasticsearchRepository" is not a really good description for a task or a problem to fix as it doesn't have a scope. Make sure your JIRA clearly states a problem. Usually we get a ticket and a PR for that ticket.

In generally we discourage adding more methods to store specific repositories as using those couples the clients to the underlying store technology, which is something a repository is introduced to prevent in the first place. Especially as we're introducing new public API, we need to perform an extra iteration of reviews to the suggested changes.

I don't quite get why you introduce a public `refresh()`method to call it yourself if the primary concern of your ticket seems to be that `index()` should call this automatically. A change, that you introduce with the very same PR.

Summing things up, I really think we should take a step back and think about the actual problem we need to solve and then apply the *minimal* amount of changes to achieve that goal. In the current set of proposed changes, we're throwing the baby out with the bathwater a bit.

xhaggi:
@olivergierke thanks for the explanation.

> I don't quite get why you introduce a public refresh()method to call it yourself if the primary concern of your ticket seems to be that index() should call this automatically.

You get me wrong, sorry for that ;) IMO it makes no sense that `index()` calls `save()` which implicit do a refresh. The `index()` method should index the document and nothing else. Therefor I've added the `refresh()` as a convenience method to easily refresh the index without the need of injecting and calling `ElasticsearchTemplate.refresh()`, where you need to pass in the entity document object.

itplayer:
@xhaggi I'm not quite understand what do you mean 

>  my problem: there is no way to use the Repository for indexing/bulk-indexing the documents without an implicit call to refresh.

If my index set every 30s to do refresh automatically and I intend to do so, why do I need a implicit refresh every time when I call `save()`. 

Every `refresh()` generate a segment in memory right? If we need a bigger segment when flush to disk, I'd like to wait 30s and let segment getting bigger.

Any idea why we still leave `refresh()` in `Save()` ? Or do we have any way to tell framework don't do `refresh()` for me implicitly?

phxql:
Is there any way to disable the automatic index refresh when saving an entity?

xhaggi:
You could only override `save()` in your custom repository or define a custom base implementation for all repositories. Take a look at 
https://docs.spring.io/spring-data/elasticsearch/docs/current/reference/html/#repositories.single-repository-behaviour

Or

https://docs.spring.io/spring-data/elasticsearch/docs/current/reference/html/#repositories.custom-behaviour-for-all-repositories

You need to use the same method signature in your custom implementation https://github.com/spring-projects/spring-data-elasticsearch/blob/771e01189ede072c4c237517893d1b252c91c4e7/src/main/java/org/springframework/data/elasticsearch/repository/support/AbstractElasticsearchRepository.java#L153

PR #123, opened 2015-10-24
Removing unnecessary type check in AbstractElasticsearchRepository

slowikps:
(Trivial change)
This check is not needed and it breaks contract. 
i.e: Stream&lt;S&gt;::iterator is valid Iterable&lt;S&gt; but will not be accepted.


mohsinh:
merged

PR #127, opened 2015-11-11
Support injecting Id in immutable beans

jloisel:
Our data-model is composed of only immutable objects. The id is not correctly injected in those objects because it needs a setter method at the moment.

The following fix injects the id directly into the field, even when the field is declared as "private final".

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
@@ -18,6 +18,7 @@ package org.springframework.data.elasticsearch.core;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -158,12 +159,18 @@ public class DefaultResultMapper extends AbstractResultMapper {
 			// Only deal with String because ES generated Ids are strings !
 			if (idProperty != null && idProperty.getType().isAssignableFrom(String.class)) {
 				Method setter = idProperty.getSetter();
-				if (setter != null) {
-					try {
+				try {
+					if (setter != null) {
 						setter.invoke(result, id);
-					} catch (Throwable t) {
-						t.printStackTrace();
+					} else if(idProperty.getField() != null){
+						final Field field = idProperty.getField();
+						if(!field.isAccessible()) {
+							field.setAccessible(true);
+						}
+						field.set(result, id);
 					}
+				} catch (Throwable t) {
+					t.printStackTrace();
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
```diff
@@ -158,12 +159,18 @@ private String buildJSONFromFields(Collection<SearchHitField> values) {
...
-						t.printStackTrace();
+					} else if(idProperty.getField() != null){
+						final Field field = idProperty.getField();
+						if(!field.isAccessible()) {
+							field.setAccessible(true);
+						}
+						field.set(result, id);
 					}
+				} catch (Throwable t) {
+					t.printStackTrace();
```
martin-g: This `t.printStackTrace()` looked bad before and still looks bad!
1) Do not catch Throwable. Catching Errors doesn't help most of the time
2) Use proper logging, not `#printStackTrace()`

src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
```diff
@@ -158,12 +159,18 @@ private String buildJSONFromFields(Collection<SearchHitField> values) {
...
-				if (setter != null) {
-					try {
+				try {
+					if (setter != null) {
 						setter.invoke(result, id);
-					} catch (Throwable t) {
-						t.printStackTrace();
+					} else if(idProperty.getField() != null){
+						final Field field = idProperty.getField();
+						if(!field.isAccessible()) {
```
martin-g: Add space after `if` to keep it consistent with the rest of the file.

src/test/java/org/springframework/data/elasticsearch/core/DefaultResultMapperTests.java
```diff
@@ -103,6 +106,42 @@ public void shouldMapGetRequestToObject() {
...
+		when(response.getId()).thenReturn(id);
+		when(response.getSourceAsString()).thenReturn(createJsonCar("Ford", "Grat"));
+
+		//When
+		ImmutableCar result = resultMapper.mapResult(response, ImmutableCar.class);
+
+		//Then
+		assertThat(result, notNullValue());
+		assertThat(result.getId(), is(id));
+		assertThat(result.getModel(), is("Grat"));
```
martin-g: Use constants (local var) for `Ford` and `Grat` as you do for `id`.
I see they are used in other test methods too - so they could be even proper constants in the test class.

src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
jloisel: It was already doing a printstacktrace, i've not changed the behavior.

martin-g: This is what I said :-)
I am not a member of the team. I'm just a user like you.
Since you make changes in this area of the code I suggest to follow the boy scout rule and improve it!

sothawo:
closing this old PR. immutable id properties are supported.

PR #129, opened 2015-12-01
DATAES-216 - Adding support to 'indices_boost' when searching against multiple indices

thiagolocatelli:
Elasticsearch java client has the support to add the clause 'indices_boost' to a query. This is useful when searching across multiple indices and, for better performance than using custom score script, a boost can be added to rank up hits from specific indices.

NativeSearchQueryBuilder does not have a "withIndicesBoost" method to allow us to do that. This pull request address this problem by adding the required methods to add such capability to the builder.

Reference: https://www.elastic.co/guide/en/elasticsearch/reference/1.4/search-request-index-boost.html

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/query/IndexBoost.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/query/IndexBoost.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/data/elasticsearch/core/query/IndexBoost.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Defines a IndexBoost to be applied on the "indices_boost" query clause
+ *
+ * @author Thiago Locatelli
+ */
+public class IndexBoost {
+
+	private String indexName;
+	private float boost;
```
martin-g: Those could/should be `final`.

src/main/java/org/springframework/data/elasticsearch/core/query/NativeSearchQuery.java
```diff
@@ -40,6 +40,7 @@
 	private List<FacetRequest> facets;
 	private List<AbstractAggregationBuilder> aggregations;
 	private HighlightBuilder.Field[] highlightFields;
+	private IndexBoost[] indicesBoost;
```
martin-g: IMO, the `List<>` member fields are more user friendly than the array/varargs ones, i.e. it is more easier at the use-side to add/remove entries.

mohsinh:
merged

PR #131, opened 2015-12-18
DATAES-224 retain newlines when reading mapping/setting files

pkraeutli:
ElasticsearchTemplate.readFileFromClasspath() reads files using BufferedReader.readLine() which skips newline characters. For JSON files this is not a problem but YAML files end up invalid because newlines are relevant.


mohsinh:
merged

PR #133, opened 2016-01-06
feature username and password for the Shield

hyeonjae:
username and password for [Shield](https://www.elastic.co/products/shield)


garpinc:
When will this pull request be integrated and into which version?

src/main/java/org/springframework/data/elasticsearch/client/TransportClientFactoryBean.java
```diff
@@ -109,6 +111,7 @@ private Settings settings() {
 				.put("client.transport.ignore_cluster_name", clientIgnoreClusterName)
 				.put("client.transport.ping_timeout", clientPingTimeout)
 				.put("client.transport.nodes_sampler_interval", clientNodesSamplerInterval)
+				.put("shield.user", String.format("%s:%s", this.username, this.password))
```
garpinc: I think that since properties can already be specified (and should be exposed as something you can set via the namespace), we can already pass arbitrary properties to plugins hence I view this as unnecessary.

PR #115, opened 2015-08-06
DATAES-188: Source filtering feature Implementation

jtsiros:
This implements source filtering to allow control over what field(s) are returned in the _source field with every hit.

See https://www.elastic.co/guide/en/elasticsearch/reference/1.6/search-request-source-filtering.html

PR #138, opened 2016-02-29
Index Partitioning

flefebure:
Hi all,

I've been digging on how to handle large ES datasets and how to partition the data in ES.
Actually, I know 2 ways to partition data in ES :
- "manually" create partition (eg one index/ month) and roll an index alias each month for the cuirrent partition (as said here by dadoonet : http://stackoverflow.com/questions/20650854/elastic-search-when-to-use-another-index)
- parameterize index name or type name in spring-data-elasticsearch, via SPel an roll the index with a Spring context bean

I tried to dig a little deeper and to evaluate if it was possible to implement some kind of partitioning closer to partitioning features offered by some SQL database (eg Oracle enterprise edition)
For theses products it's possible to define one or several partition keys that target column of the partitioned table. The data are  then inserted in the good partition depending of theses columns values.

So I tried to build such a system on top of the 1.3.2.RELEASE version. (Please note that this current branch "performance-feature" merges also https://github.com/spring-projects/spring-data-elasticsearch/pull/115  (ability to do source filtering))

Use of this feature is illustrated in testsU (https://github.com/flefebure/spring-data-elasticsearch/tree/performance-features/src/test/java/org/springframework/data/elasticsearch/core/partition)

In summary :
- You have to drop 2 new beans on the spring context :
  
  `<bean name="elasticsearchPartitionsCache"
      class="o.s.d.e.c.partition.DefaultElasticsearchPartitionsCache">
      <constructor-arg name="client" ref="client"/>
      <property name="elasticsearchOperations" ref="elasticsearchTemplate"/>
   </bean>`
  
  `<bean name="elasticsearchPartitioner"
       class="o.s.d.e.c.partition.DefaultElasticsearchPartitioner">
      <property name="elasticsearchPartitionsCache" ref="elasticsearchPartitionsCache"/>
      <property name="elasticsearchOperations" ref="elasticsearchTemplate"/>
  </bean>`
  
  `<bean name="elasticsearchTemplate"
        class="o.s.d.e.c.ElasticsearchTemplate">
      <constructor-arg name="client" ref="client"/>
      <property name="elasticsearchPartitioner" ref="elasticsearchPartitioner"/>
  </bean>`
  - Document annotation is enriched with Partition methods :

`@Document(indexName = "index", type = "dateparttype", indexStoreType = "memory", shards = 1,
        replicas = 0, refreshInterval = "-1", partitionersFields = "creationDate", partitioners = Partitioner.date_range, partitionersParameters = "YYYYMM")
public class DatePartitionedEntity {`
- You can also define multi-level partitions (ex partitions for a customer then a period)
- Partitions, which are ES indexes are created on-the-fly at index time. Theses index names contains the partition start boundary (eg "myindex-2001601")
- document ID are also altered on-the-fly to contain the partition start boundary. This does permit to target the good partition when we get or update a existing document.
- For each search query, the ES cluster is interrogated to determine a list of partitions for a given document type. It is possible to cache this list if you drop an "esPartitions" ehcache on the classpath

`@Cacheable(value = "esPartitions")
    public List<String> listIndicesForPrefix(String prefix) {}`

`@Cacheable(value = "esPartitions")
    public List<String> listTypesForPartition(String partition) {}`
- If you use the template as usual, a full scan of every existing partitions will be done
- If you want to target partitions, you can enrich the NativeSearchQuery or the CriteriaQuery

`SearchQuery query = new NativeSearchQueryBuilder().withQuery(QueryBuilders.matchAllQuery()).withPartition(new LongPartition("id", 1000, 2000)).build();
        list = template.queryForList(query, LongPartitionedEntity.class);` 
- if you Spring ES repositories as usual, partitions full scan will be triggered. If you want to target a partition range, you can also enrich the repository methods with a "Partition" parameter

`public interface StringPartitionedEntityRepository extends ElasticsearchRepository<StringPartitionedEntity, String> {
    public List<StringPartitionedEntity> findByCustomerId(String customerId, Partition partition, Pageable pageable);
}`

Well, I'm not really new to java but quite new to open source and Spring contributions. I don't really know if my work meets the Spring requirements, but if some Spring folks think it's interesting I can try to consolidate my work to follow their guidelines.
Eventually, if my work breaks too much the library design, I think nevertheless that the feature is interesting (I do'nt understand why ES folks think about it) and could be implemented another way.

Any critic or remark will be appreciated

Franck


pivotal-issuemaster:
@flefebure Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=138) to manually synchronize the status of this Pull Request.

sothawo:
closing this; this surely was an interesting idea at the time this PR was created, but alas nobody at that time care about it.
I think that today this should be adressed by either using SpEL based index names which are working, and by combining this with index emplates - which are not yet integrated into Spring Data Elasticsearch

PR #139, opened 2016-03-04
DATAES-238: ElasticsearchTemplate.prepareUpdate() does not preserve routing parameter which is required for updating child documents

saigyo:
When updating a child document in a parent-child document relationship, Elasticsearch requires a routing parameter on the UpdateRequest.

ElasticsearchTemplate.update(), respectively ElasticsearchTemplate.prepareUpdate() does not preserve the routing parameter specified in the the UpdateRequest (wrapped in the UpdateQuery parameter).

The fix is a one-liner in ElasticsearchTemplate.update().


mohsinh:
Thanks for the PR, merged

andrea-patricelli:
Hi,

can you explain me WHY this fix has only be scheduled (and now applied) for 2.0?? And not for an intermediate minor version?
I'm using 1.3.4.RELEASE with ES 1.7.4 and so I'm obliged to upgrade to 2.0. Sooner I'll do this, but I think that 1.x (like me) users must not be blocked by such impacting bug.

yangxinbo:
Spring-data-Elasticsearch relationship between grandparents and grandchildren how to define? Parent-child is possible

PR #141, opened 2016-03-15
DATAES-234 - Use Set to store qualifiers in CDI extension.

mp911de:
Qualifiers are now stored in their original set instead of using the toString representation.


mohsinh:
Thanks, merged

PR #151, opened 2016-06-27
[DATAES-270] Add new methods for query response scrolling to the Elas…

gzsombor:
Add new methods for query response scrolling to the ElasticsearchRepository, so no need to directly access the ElasticsearchTemplate for the client code


sothawo:
closing this as Repsoitories support `Stream<T>` so there is no need for this.

PR #154, opened 2016-07-26
DATAES-260 - StacOverflow when two reverse relationship.

xinevic:
The problem was in incorrect ignore list:
correct ignore list is in parrent field annotation, but earlier field name was compared with current field ignore list.

Fixed: comparing current field name with ignore list from parent field.


akonczak:
Thank you for you contribution, the changes has been rebased and merged to the master.

PR #155, opened 2016-08-07
DATAES-228 Fix id handling for indexing

praseodym:
When an id is set in the IndexQuery object, use it without requiring an
@Document annotation. When it is not, try to resolve it from the object
when it has @Document and @Id annotations.


akonczak:
Thank you very much for your PR. Changes has been merged to master.

PR #156, opened 2016-08-18
Fixed DATAES-281 Can't save entity without id setter

hyysguyang:
Fixed DATAES-281

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
@@ -170,17 +170,11 @@ public class DefaultResultMapper extends AbstractResultMapper {
 
 	private <T> void setPersistentEntityId(T result, String id, Class<T> clazz) {
 		if (mappingContext != null && clazz.isAnnotationPresent(Document.class)) {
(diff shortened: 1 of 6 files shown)


olivergierke:
I'd want to veto this. The `PersistentPropertyAccessor` API is in place to avoid Spring Data modules having to deal with such low level decisions. Let alone the fact that the exception handling is quite a no-no. `PersistentPropertyAccessor` honors the configuration you have in place when obtaining the values, so I am wondering what the actual problem is in the first place.

Let's continue the discussion about what the actual problem is on the ticket before we thing about the way we implement it.

olivergierke:
Oh damnit. I totally misread your PR (sometimes red and green are hard to distinguish 🙄). Totally my bad. Actually it looks like `DefaultResultMapper` might need a general overhaul as it's also doing a lot of reflection based stuff that should rather be achieved through our `PersistentEntity` API.

hyysguyang:
@olivergierke  Sorry, maybe I have not described it clean.   We can use the `PersistentPropertyAccessor` to set/get the entity id, and then honor the configuration of `usePropertyAccess`

This PR is just fixing the setter issue, if it's OK, then we also need to update `ElasticsearchTemplate.getPersistentEntityId` to use the `PersistentPropertyAccessor`.

The `DefaultResultMapper` seems OK, just the `setPersistentEntityId` use the Reflection.

 The detail description is here [DATAES-281)](https://jira.spring.io/browse/DATAES-281) .

PR #158, opened 2016-11-10
DATAES-309: Documentation updates

fhopf:


sothawo:
I will integrate this into DATAES-497

sothawo:
included in #291

PR #165, opened 2016-12-20
DATAES-287 - added new date formats for ES>2.0 missing in DateFormat

javamachr:
ES 2.0 changed the way they handle dates and dateFormats.
This pull added missing date formats based on 2.4 doc.

It also added test demonstrating date format mapping issue described in [DATAES-287](https://jira.spring.io/browse/DATAES-287). The test is actually checking the wrong value so it passes while it should not pass if the mapping worked correctly.

The issue with ignored DateFormat is much more complicated - simple workaround can be achieved adding @JsonFormat annotation to particular field as described in [DATAES-287](https://jira.spring.io/browse/DATAES-287).

I have signed and agree to the terms of the SpringSource Individual
Contributor License Agreement.


dkirrane:
Never merged?

sothawo:
We intend to remove Jackson usage entirely and use convention-based mapping on top of Spring Data's entity metadata.

PR #172, opened 2017-02-23
Added putMapping implementations to allow for custom indexName and type

carofe82:
This new putMapping implementations allow you to use a custom indexName
and/or type and still use the Field annotations for the mapping

Please let me know if you like this idea. Thank you.


carofe82:
There is a small bug in the putMapping(indexName, type, mapping). I'll send the fix with a refined code if you guys are interested in the idea.

PR #177, opened 2017-03-15
DATAES-332 - Added index type parameter to GetQuery object

MateuszPu:
https://jira.spring.io/browse/DATAES-332


xhaggi:
Starting with version 6.x of Elasticsearch, type support is limited to a single type and with 7.x no longer supported.

https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html

PR #180, opened 2017-03-28
DATAES-317 - Introduce query logging in ElasticsearchTemplate.

xhaggi:
The java client of elasticsearch does not provide an easy way to log the serialized query (JSON) by setting the log level of some logger to DEBUG. Therefore this PR adds logging capability to the ElasticsearchTemplate because it enclose the ES `SearchRequestBuilder` where you need to call `toString()` to get the real JSON query.

The logging is done by a separate logger `org.springframework.data.elasticsearch.core.QUERY` which needs to be set to log level DEBUG to output the serialized query.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
@@ -19,7 +19,7 @@ import static org.elasticsearch.client.Requests.*;
 import static org.elasticsearch.index.VersionType.*;
 import static org.elasticsearch.index.query.QueryBuilders.*;
 import static org.springframework.data.elasticsearch.core.MappingBuilder.*;
-import static org.springframework.util.CollectionUtils.isEmpty;
+import static org.springframework.util.CollectionUtils.*;
 
 import java.io.BufferedReader;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
```diff
@@ -931,7 +930,13 @@ private SearchResponse doSearch(SearchRequestBuilder searchRequest, SearchQuery
 				searchRequest.addAggregation(aggregatedFacet.getFacet());
 			}
 		}
-		return getSearchResponse(searchRequest.setQuery(searchQuery.getQuery()).execute());
+		return getSearchResponse(searchRequest.setQuery(searchQuery.getQuery()));
+	}
+
+	private SearchResponse getSearchResponse(SearchRequestBuilder requestBuilder) {
+		if (QUERY_LOGGER.isDebugEnabled())
+			QUERY_LOGGER.debug(requestBuilder.toString());
```
shawnjohnson: I'm not sure if Spring has any guidelines for this, but I found this old document recommending to include curly-braces around if statements - http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-142311.html

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
```diff
@@ -136,7 +137,8 @@
  */
 public class ElasticsearchTemplate implements ElasticsearchOperations, ApplicationContextAware {
 
-	private static final Logger logger = LoggerFactory.getLogger(ElasticsearchTemplate.class);
+	private static final Logger QUERY_LOGGER = LoggerFactory.getLogger("org.springframework.data.elasticsearch.core.QUERY");
```
shawnjohnson: I like this idea, though I don't know if there are any conventions within Spring / Spring Data to look towards following for this type of logging?  For example in the JPA project they have the following options for logging queries - http://www.baeldung.com/sql-logging-spring-boot

olivergierke:
I've polished and merged that for Lovelace and Kay.

xhaggi:
@olivergierke thanks 👍

PR #184, opened 2017-07-05
DATAES-233 Utility class to support rolling index strategy

zzt93:

### DIFF
--- a/src/test/java/org/springframework/data/elasticsearch/entities/Chat.java
+++ b/src/test/java/org/springframework/data/elasticsearch/entities/Chat.java
@@ -30,7 +30,7 @@ import org.springframework.data.elasticsearch.annotations.Documents;
 @Getter
 @NoArgsConstructor
 @Builder
-@Documents(indexPattern = "chat_#{indexNameVar}", type = "chat")
+@Documents(indexPattern = "chat_#{indexNameVar.toString()}", type = "chat")
 public class Chat {
 
 	@Id private String id;
--- a/src/test/java/org/springframework/data/elasticsearch/entities/Chat.java
+++ b/src/test/java/org/springframework/data/elasticsearch/entities/Chat.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.entities;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import org.springframework.data.annotation.Id;
+import org.springframework.data.elasticsearch.annotations.Documents;
+
+/**
+ * @author zzt
+ */
+@Setter
+@Getter
+@NoArgsConstructor
+@Builder
+@Documents(indexPattern = "chat_#{indexNameVar}", type = "chat")
+public class Chat {
+
+	@Id private String id;
+	private String sender;
+	private String receiver;
+	private String content;
+
+	public Chat(String id, String sender, String receiver, String content) {
+		this.id = id;
+		this.sender = sender;
+		this.receiver = receiver;
+		this.content = content;
+	}
+}
--- a/src/test/java/org/springframework/data/elasticsearch/repository/support/IndexNameVarTest.java
+++ b/src/test/java/org/springframework/data/elasticsearch/repository/support/IndexNameVarTest.java
@@ -57,7 +57,7 @@ public class IndexNameVarTest {
 		elasticsearchTemplate.refresh(Chat.class);
 	}
 
-	@Test
+	@Test // DATAES-233
 	public void setVars() throws Exception {
 		// given
 		indexNameVar.setVars(first);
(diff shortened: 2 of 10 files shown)


edwardcapriolo:
Wouldnt using aliases accomplish the same thing?

zzt93:
It seems not same. 

- Alias use different names point to a single index;
- Index sharding using same clazz with suffix to point to multiple indicies, to use different index at different situation;

zzt93:
It seems not same. 

- Alias use different names point to a single index;
- Index sharding using same clazz with suffix to point to multiple indicies, to use different index at different situation;

edwardcapriolo:
The complex part will be how to decide what index to writer to and read from fo a given search.You end up trying to build a table-by-day like soluttion

src/test/java/org/springframework/data/elasticsearch/repository/support/IndexNameVarTest.java
```diff
@@ -0,0 +1,91 @@
...
+		indexNameVar.setVars(second);
+		repository.save(new Chat("10", "s1", "s2", "what are you doing"));
+		repository.save(new Chat("11", "s2", "s1", "having meal"));
+		Callable<Void> r1 = () -> {
+			// when
+			indexNameVar.setVars(first);
+			Page<Chat> s1 = repository.findBySender("s1", PageRequest.of(0, 10));
+			// then
+			assertThat(s1.getTotalElements() == 2, is(true));
+			System.out.println(s1.getContent());
```
edwardcapriolo: println here

src/test/java/org/springframework/data/elasticsearch/repository/support/IndexNameVarTest.java
```diff
@@ -0,0 +1,91 @@
...
+			return null;
+		};
+		Callable<Void> r2 = () -> {
+			// when
+			indexNameVar.setVars(second);
+			Page<Chat> s1 = repository.findBySender("s1", PageRequest.of(0, 10));
+			// then
+			assertThat(s1.getTotalElements() == 1, is(true));
+			System.out.println(s1.getContent());
```
edwardcapriolo: println here

src/test/java/org/springframework/data/elasticsearch/entities/Chat.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.data.elasticsearch.annotations.Documents;
+
+/**
+ * @author zzt
+ */
+@Setter
+@Getter
+@NoArgsConstructor
+@Builder
+@Documents(indexPattern = "chat_#{indexNameVar.toString()}", type = "chat")
```
edwardcapriolo: Can this reference a field in the document that contains date information?It feels like the reference should be to something in the object to be persisted

edwardcapriolo: @Field(name="createDate" store=false index= false)
private Date createDate

or @IndexVar 
private Date createDate

edwardcapriolo:
Can one of the contributers comment on the approach and the code. table by day is very popular and if there is no support many of my applications can not use spring-data-elastic

zzt93:
@edwardcapriolo This PR is opened two years ago, I will try to recall the thought and consider your suggestions to adjust the implementation.

sothawo:
I just wrote a [comment in the jira issue](https://jira.spring.io/browse/DATAES-233), I would like to first challenge this solution, if it the best way to solve a problem like rolling indices.

zzt93:
@sothawo I read your comment in jira. As far as I can see, the best practice is to put the [index template](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html) in ES at the beginning, so we don't need to do any check for index existence and let ES do other thing. Then, we can just save entity to ES easily. 

As far as about the index name binding mechanism, @edwardcapriolo provides a suggestions: using a field in that entity, which is more intuitive. In that case, we can't use a global bean for root of Spring EL. We need to set the root object (Again, using thread local context to parse index name) before saving this kind of document, however, in that case, It seems only making index obj a little easier, and not for all other operation. Besides, it break the backward compatibility, so I wonder whether it is a better solution.

@sothawo How do think?

sothawo:
why a ThreadLocal? What if some process submits tasks to a threadpool that insert into ES? Then these tasks run on different threads but should probably use the same index, although there are different name reolvers?

Having the option to define a field on the object to provide the indexname - perhaps together with a `@Transient` annotation to not store it in the index - might be a solution which could override the indexname set in the `@Document` annotation. Then the indexname creation is completely out of Spring Data Elasticsearch and up to the code creating the objects to be stored in the index.

But we need to have control over the creation of the index, because if we let Elasticsearch just create an index and store some document in it, then Elasticsearch will use the default mapping for the properties of the document, and we loose all the information that might be annotated with `@Field` or the other relevant configurations that must be put into the mapping.

zzt93:
Maybe I not express clearly, but `ThreadLocal` is not the core of index suffix. The suffix must be decided by biz logic (may be time or id etc), then use a threadlocal bean to store and parse. ThreadLocal is just to avoid thread issue in a web service, any other way to solve concurrent bean access is fine.

When it comes to index creation, **index template** is more convenient and suitable for this kind index partitioning as its [official guide said](https://www.elastic.co/guide/en/elasticsearch/guide/current/index-templates.html). We can use `@Field` and relevant conf to create right index template, rather than manually create index.

sothawo:
The creation of a name for the index is business logic, that's right. The question is, how is the index name for the current document passed into Spring Data Elasticsearch? 

This can be for example by a bean resolved via SpEL. If this is not feasible, we could add an annotation like `@IndexName` which can be put on a property of the document designating the value of this property being the index name.

But there is the problem that for new indices, the `ElasticsearchOperations.createIndex()` and ` ElasticsearchOperations.putMapping()` methods need to be called to have a consistent mapping across the different indices. Applications using `ElasticsearchOperations` implementations (`*Template` classes) directly, they can handle this by themselves. 

But when the `*Repository` classes are used, this must be handled be the Repository implementations and this is currently not supported. (1)

A different solution would be to support *index templates* like you mentioned, but this is not implemented yet. (2)

I will create two tickets later for (a) and (2)

sothawo:
DATAES-233 is closed as we now support index templates, and have an improved alias management. For how to get an index name dynamically from an entity I wrote a short example (https://www.sothawo.com/2020/08/use-an-index-name-defined-by-the-entity-to-store-data-in-spring-data-elasticsearch-4-0/).

PR #186, opened 2017-08-02
DATAES-169 Geo shape support

hefay:
Basic support for geo_shape types and Criteria queries

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,17 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- required by elasticsearch for geo_shape support -->
+        <dependency>
+            <groupId>com.vividsolutions</groupId>
+            <artifactId>jts</artifactId>
+            <version>1.13</version>
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.spatial4j</groupId>
+            <artifactId>spatial4j</artifactId>
+            <version>0.6</version>
+        </dependency>
 
 
     </dependencies>
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/GeoShapeField.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/GeoShapeField.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * @author Lukas Vorisek
+ * @since 3.0
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@Documented
+public @interface GeoShapeField {
+	GeoShapeTree tree() default GeoShapeTree.geohash;
+	String precision() default "50m";
+	GeoShapeOrientation orientation() default GeoShapeOrientation.ccw;
+}
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/GeoShapeOrientation.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/GeoShapeOrientation.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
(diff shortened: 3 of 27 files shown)


fer-marino:
any news on this? it is quite an important feature...

MoMenne:
+1

fguillotpro:
looking forward to seeing this merged

pivotal-issuemaster:
@hefay Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=186) to manually synchronize the status of this Pull Request.

sothawo:
@mp911de @christophstrobl for a feature like this, should we add the dependency on two additional libraries (com.vividsolutions.jts is from December 2012, org.locationtech.spatial4j from December 2017)?

mp911de:
From a quick glance, we seem to use little from these libraries. Can you elaborate a bit why these are required and why we cannot reuse geo classes from Spring Data Commons @sothawo @hefay?

hefay:
Hello. 

> From a quick glance, we seem to use little from these libraries. Can you elaborate a bit why these are required and why we cannot reuse geo classes from Spring Data Commons @sothawo @hefay?

Hello. It's long time ago when I tried to do this change and lot of stuff has been changes since that. I will try take a look again at that and update this change.

pivotal-issuemaster:
@hefay Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=186)!

hefay:
@mp911de These libraries (JTS, spatial4j) was added because ES is not able to do spatial queries without them. So I wanted to provide them for the user. Should I include them only for tests and let the user add them itself?

https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/java-geo-queries.html

sothawo:
Thanks for the link to the ES documentation. Making the dependencies optional is not enough, because the code in the MappingBuilder / EntityMapper etc needs these classes. So we need to find a way to encapsulate this logic so that it is only active/used when the libraries are on the classpath, otherwise we run into class not found exceptions.

As we are currently reworking the internal object mapping code, it might be the best thing to keep this PR on hold until we finished that; if you would now try to resolve the conflicts, there would be needed changes after our refactoring as well.

So for the time being, we'll leave this PR, and then later - probably in October - we will check how we can integrate this.

hefay:
@sothawo Ok. If that is the case it seems better to wait for your changes. After that, I would like to create this PR again.

thlaegler:
Hi team, thanks for your great work!!! 👍

Are there any news on this? How far has the refactoring of the internal object mapping gone? I'm really desperately waiting to see this feature.

A "quick-win" (at least for me) would be to have `geo_shape` in the `FieldType` enum or to have a `@GeoShape` annotation so that at least the index mapping would have the right field type (`geo_shape`) when getting created by spring (e.g. similar like `@GeoPointField` annotation). Is it worth to submit a PR just for that "quick-win".

sothawo:
Hello Thomas, thanks for reminding about this topic - I forgot to check the old PRs regularly.

Going over the issue and the PR: that's definitely something worth to have. I don't think that we can merge this PR straight away as it has too many conflicts with the currrent master after all our refactoring. But it would be a pity to loose the work that has been done.

A first and easiest step is to add a `FieldType.geo_shape` so that the index mappings are written - although I think, a `GeoShape` annotation would be better as we need to pass in the  different mapping options as arguments to the annotation. This would have to be processed in the `MappingBuilder` class (feel free to create a [Jira issue](https://jira.spring.io/projects/DATAES/issues) for that). 

The next step would be to add `Converter` implementations that can map from/to the Geo Types to a `Map<String, Object>`. These should only be registered when the corresponding libraries are available on the classpath, but that's not a problem.

Once we have the mappings and the converters we can store and retrieve the data.

The last step would be to implement _geo_shape_ queries.

To wrap it up, I'd prefer to split this into separate tickets, but take what is possible from this PR - especially crediting the author of this PR, and implement it. If you have further questions, don't hesitate to ask

sothawo:
So after the reease of version 4.0 and preparing 4.1 I finally find the time to have a deeper look at this. As the PR targeted 3.0 where we had no support for the REST client, and now in 4.0 the transport client is deprecated (and some of the Java functions as well https://www.elastic.co/guide/en/elasticsearch/client/java-api/7.9/java-geo-queries.html) I have to check what we can use and how.

I will probably take this code and split it into different parts (the mappings, the converter, the queries) and integrate these into the current code.

I will of course keep the author information when copying the code from your repository

hefay:
Hi. @sothawo please let me know if I can help you at some point. I haven't much time these days, but maybe I can try at least clarify some potential confusion in my original PR.

sothawo:
After finishing the three new issues [DATAES-929](https://jira.spring.io/browse/DATAES-929), [DATAES-930](https://jira.spring.io/browse/DATAES-930) and  [DATAES-931](https://jira.spring.io/browse/DATAES-931) this is finally done.

Thanks a lot one more time for submitting this PR in the first place.

I could take the part for the index mappings with the annotation, adapting it to the newest version of Elasticsearch.

The storing/retrieving, i.e. mapping of the objects needed a rewrite as we don't use Jackson anymore for this.

The part for building the `CriteriaQuery` is a complete new part, I have a solution where the filter query is sent as a wrapper query built from the GeoJson string. By this, we don't need all the builders from Elasticsearch and we don't need to add the dependencies on the _jts_ and _spatial4j_ libraries.

But even if I did not use large parts of your PR, it gave me quite some ideas what to do.

PR #187, opened 2017-08-31
DATAES-386 fix NullPointerException in FacetedPageImpl

juliocamarero:
This bug can not be reproduced in the latest master since the FacetedPageImpl class has been refactored. It does affect the branch 2.2.x
When calling hasFacets or getFacets in any Page that extends FacetedPage a NullPointerException will be thrown.
The reason is that facets can be null but the logic inside processAggregations method is asuming the variable is always initialized.

- [x]  You have read the Spring Data contribution guidelines.
- [x]   There is a ticket in the bug tracker for the project in our JIRA.
- [x]   You use the code formatters provided here and have them applied to your changes. Don’t submit any formatting related changes.
- [x]   You submit test cases (unit or integration tests) that back your changes.
- [x]   You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/FacetedPageImpl.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/FacetedPageImpl.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.elasticsearch.search.aggregations.Aggregation;
+import org.elasticsearch.search.aggregations.Aggregations;
 import org.elasticsearch.search.aggregations.bucket.histogram.Histogram;
 import org.elasticsearch.search.aggregations.bucket.range.Range;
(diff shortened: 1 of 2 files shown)


juliocamarero:
Hi @olivergierke!

I’ve just seen that you have closed this pull request without adding any information. Could you please give me some feedback on what I did wrong so that I can improve it?

Thank you!

olivergierke:
The close was triggered by us removing the 2.2.x branch as it was reserved for a potential intermediate release train between Ingalls (2.1) and Kay (3.0). Feel free to rebase the PR onto Ingalls as it should be virtually identical to 2.2 and re-submit. 🙃

PR #192, opened 2017-10-28
3.0.x

lives78:


michael-wirth:
I wonder why building new releases if there are plenty on major bugs still open and nobody of the core team is taking care of the bugs and pull requests? as a matter of fact nobody from the core team seems to online since July and don't react on any emails, which let's me think thus project is abanded. so the question is why support an abandoned project?

PR #199, opened 2018-04-09
[DATAES-438] get rid of deprecation warning because of invalid value for option 'index' on id field mapping

xhaggi:
There is a left-over `not_analysed` index option mapping after the upgrade to 5.x.
This PR address the issue and replace it with `true`.


xhaggi:
@akonczak please backport this PR to 3.0.x

akonczak:
done - but I'm afraid we missed out release train today ;/

PR #207, opened 2018-06-13
DATAES-462 - Add score mapping support to DefaultResultMapper

xhaggi:
See https://jira.spring.io/browse/DATAES-462

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Score.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Score.java
@@ -1,0 +1,23 @@
+package org.springframework.data.elasticsearch.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.data.annotation.ReadOnlyProperty;
+
+/**
+ * Specifies that this field is used for storing the document score.
+ * 
+ * @author Sascha Woo
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@Documented
+@Inherited
+@ReadOnlyProperty
+public @interface Score {
+}
--- a/src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
@@ -53,6 +53,7 @@ import com.fasterxml.jackson.core.JsonGenerator;
  * @author Chris White
  * @author Mark Paluch
  * @author Ilkang Na
+ * @author Sascha Woo
  */
 public class DefaultResultMapper extends AbstractResultMapper {
 
@@ -62,7 +63,8 @@ public class DefaultResultMapper extends AbstractResultMapper {
 		super(new DefaultEntityMapper());
 	}
 
-	public DefaultResultMapper(MappingContext<? extends ElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty> mappingContext) {
+	public DefaultResultMapper(
+			MappingContext<? extends ElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty> mappingContext) {
 		super(new DefaultEntityMapper());
 		this.mappingContext = mappingContext;
 	}
@@ -81,6 +83,8 @@ public class DefaultResultMapper extends AbstractResultMapper {
 	@Override
 	public <T> AggregatedPage<T> mapResults(SearchResponse response, Class<T> clazz, Pageable pageable) {
 		long totalHits = response.getHits().getTotalHits();
+		float maxScore = response.getHits().getMaxScore();
+
 		List<T> results = new ArrayList<>();
 		for (SearchHit hit : response.getHits()) {
 			if (hit != null) {
@@ -90,14 +94,17 @@ public class DefaultResultMapper extends AbstractResultMapper {
 				} else {
 					result = mapEntity(hit.getFields().values(), clazz);
 				}
+
 				setPersistentEntityId(result, hit.getId(), clazz);
 				setPersistentEntityVersion(result, hit.getVersion(), clazz);
+				setPersistentEntityScore(result, hit.getScore(), clazz);
 				populateScriptFields(result, hit);
 				results.add(result);
 			}
 		}
 
-        return new AggregatedPageImpl<T>(results, pageable, totalHits, response.getAggregations(), response.getScrollId());
+		return new AggregatedPageImpl<T>(results, pageable, totalHits, response.getAggregations(), response.getScrollId(),
+				maxScore);
 	}
 
 	private <T> void populateScriptFields(T result, SearchHit hit) {
@@ -112,8 +119,8 @@ public class DefaultResultMapper extends AbstractResultMapper {
 						try {
 							field.set(result, searchHitField.getValue());
 						} catch (IllegalArgumentException e) {
-							throw new ElasticsearchException("failed to set scripted field: " + name + " with value: "
-									+ searchHitField.getValue(), e);
+							throw new ElasticsearchException(
+									"failed to set scripted field: " + name + " with value: " + searchHitField.getValue(), e);
 						} catch (IllegalAccessException e) {
 							throw new ElasticsearchException("failed to access scripted field: " + name, e);
 						}
@@ -177,9 +184,7 @@ public class DefaultResultMapper extends AbstractResultMapper {
 	}
 
 	private <T> void setPersistentEntityId(T result, String id, Class<T> clazz) {
-
 		if (mappingContext != null && clazz.isAnnotationPresent(Document.class)) {
-
 			ElasticsearchPersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(clazz);
 			ElasticsearchPersistentProperty idProperty = persistentEntity.getIdProperty();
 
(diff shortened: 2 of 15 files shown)


src/main/java/org/springframework/data/elasticsearch/core/ScoredEntity.java
```diff
@@ -0,0 +1,12 @@
+package org.springframework.data.elasticsearch.core;
+
+/**
+ * @author Sascha Woo
+ */
+public interface ScoredEntity {
+
+	float getScore();
+
+	void setScore(float score);
```
olivergierke: Is that an interface that's supposed to be implemented by user code? If so, does it makes sense that the score can be modified? That feels quite invasive. Should we provide an annotation based approach instead, so that users can simply annotate a field in their domain class and we invoke this reflectively (via `PersistentEntity.getPropertyAccessor(…).setProperty(scoreProperty, value);` in `DefaultResultMapper? I guess, that would need `ElasticsearchPersistentEntity` to expose a score property.

src/main/java/org/springframework/data/elasticsearch/core/ScoredPage.java
```diff
@@ -0,0 +1,13 @@
+
+package org.springframework.data.elasticsearch.core;
+
+import org.springframework.data.domain.Page;
+
+/**
+ * @author Sascha Woo
+ */
+public interface ScoredPage<T> extends Page<T> {
```
olivergierke: A bit of Javadoc highly appreciated :).

src/main/java/org/springframework/data/elasticsearch/core/query/AbstractQuery.java
```diff
@@ -160,4 +164,13 @@ public IndicesOptions getIndicesOptions() {
 	public void setIndicesOptions(IndicesOptions indicesOptions) {
 		this.indicesOptions = indicesOptions;
 	}
+
+	@Override
+	public boolean getTrackScores() {
+		return trackScores;
+	}
+
+	public void setTrackScores(boolean trackScores) {
```
olivergierke: At a quick glance, this setter doesn't seem to be used anywhere, right?

olivergierke: Oh, it is. From `NativeSearchQueryBuilder`. However, it should be in package scope then to limit the API surface.

src/main/java/org/springframework/data/elasticsearch/core/ScoredEntity.java
olivergierke: After talking to @christophstrobl, it looks like the Solr module has an [`@Score`](https://github.com/spring-projects/spring-data-solr/blob/master/src/main/java/org/springframework/data/solr/repository/Score.java) annotation that's a `@ReadOnlyProperty`. Would be cool if you could follow that design. A good starting point for the integration into the metamodel is probably [`SolrPersistentEntity`](https://github.com/spring-projects/spring-data-solr/blob/fd61a1672ac5508a80ba68db76fdd288d3e3f081/src/main/java/org/springframework/data/solr/core/mapping/SolrPersistentEntity.java#L36-L53).

xhaggi: Thought exactly the same before I saw this comment ;) Will address your comments

olivergierke: Feel free to ping me in [Gitter](https://gitter.im/spring-projects/spring-data) if you need help regarding the meta-model and property access.

src/main/java/org/springframework/data/elasticsearch/annotations/Score.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Specifies that this field is used for storing the document score.
+ * 
+ * @author Sascha Woo
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@Documented
+@Inherited
+public @interface Score {
```
olivergierke: Let's meta-annotate this with `@ReadOnlyProperty` so that `@Transient` is not needed in the domain model as we always consider a score property metadata, right?

olivergierke:
Let me take over quickly. I'll add a commit that takes care of proper integration with the `MappingContext`.

src/main/java/org/springframework/data/elasticsearch/core/query/AbstractQuery.java
manuelsx28: This setter needs to be public for cases where `NativeSearchQueryBuilder` is not used.

olivergierke: I am probably too unfamiliar with the ES specifics here. I am merely looking at this from an "How much API do we expose?" point of view and API that assumes or exposes mutability is usually candidate for extra scrutiny.

xhaggi:
@olivergierke force-pushed with everything addressed. Now it use the mapping context + meta model to determine the score property etc.

src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
```diff
@@ -206,4 +209,16 @@ private String buildJSONFromFields(Collection<DocumentField> values) {
 			}
 		}
 	}
+
+	private <T> void setPersistentEntityScore(T result, float score, Class<T> clazz) {
+		if (mappingContext != null && clazz.isAnnotationPresent(Document.class)) {
```
olivergierke: Do we need that guard? Shouldn't the constructor of `DefaultResultMapper` already make sure that it's never `null`? What's the second guard forI'd vote to avoid any Java-native reflection in the context of actually working with the metamodel.

src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
```diff
@@ -206,4 +209,16 @@ private String buildJSONFromFields(Collection<DocumentField> values) {
...
+			ElasticsearchPersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(clazz);
+			ElasticsearchPersistentProperty scoreProperty = persistentEntity.getScoreProperty();
```
olivergierke: Shall we use `….hasScoreProperty()` here to avoid the downstream `null` check?

src/main/java/org/springframework/data/elasticsearch/core/DefaultResultMapper.java
```diff
@@ -206,4 +209,16 @@ private String buildJSONFromFields(Collection<DocumentField> values) {
...
+			Class<?> type = scoreProperty.getType();
```
olivergierke: Does it make sense to already reject score properties with other types than `float`/`Float` when the `ElasticsearchPersistentProperty` is created? That would fail much faster and not let the application bootstrap and the score not being mapped without any indication that the type doesn't match.

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
```diff
@@ -49,4 +51,22 @@
 	String settingPath();
 
 	boolean isCreateIndexAndMapping();
+
+	/**
+	 * Returns whether the {@link ElasticsearchPersistentEntity} has an score property. If this call returns
+	 * {@literal true}, {@link #getScoreProperty()} will return a non-{@literal null} value.
+	 *
+	 * @return false when {@link ElasticsearchPersistentEntity} does not define a score property.
```
olivergierke: Let's add `@since 3.1` for newly introduced API. Probably also for the method below, and `ScoredPage` etc.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -165,7 +179,21 @@ public void addPersistentProperty(ElasticsearchPersistentProperty property) {
...
+		if (property.isScoreProperty()) {
+			ElasticsearchPersistentProperty scoreProperty = this.scoreProperty;
+
+			if (scoreProperty != null) {
+				throw new MappingException(
+						String.format("Attempt to add score property %s but already have property %s registered "
+								+ "as version. Check your mapping configuration!", property.getField(), scoreProperty.getField()));
+			}
+
+			Assert.isTrue(property.getType() == Float.class || property.getType() == Float.TYPE, "Score property must be of type float!");
```
olivergierke: I'd move that check into the creation of the `ElasticsearchPersistentProperty`.

src/main/java/org/springframework/data/elasticsearch/annotations/Score.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Specifies that this field is used for storing the document score.
+ * 
+ * @author Sascha Woo
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@Documented
+@Inherited
+@ReadOnlyProperty
```
olivergierke: I've just checked the `EntityWriter` implementation and it looks like it

src/main/java/org/springframework/data/elasticsearch/core/query/AbstractQuery.java
manuelsx28: There are cases in which the query is created using the constructors of `NativeSearchQuery` and then additional setters are called. If we want to enforce the use of the `NativeSearchQueryBuilder`, all constructors and setters should be package scoped. But we should not do that, as it would break existing code. So in my opinion, this setter should be public analog to the other setters. In our internal projects, we implemented a less complex search query builder implementation which needs to have access to the `setTrackScores` method.

olivergierke: Understood, thanks. Looks like Sascha opened up the visibility to `public` again.

src/main/java/org/springframework/data/elasticsearch/annotations/Score.java
olivergierke: I've just filed [DATAES-464](https://jira.spring.io/browse/DATAES-464) and am going to push a fix for that. With that in place, the read-only properties shouldn't be written to Elasticsearch.

src/main/java/org/springframework/data/elasticsearch/core/query/AbstractQuery.java
xhaggi: Yeah we discussed it before.

src/main/java/org/springframework/data/elasticsearch/annotations/Score.java
xhaggi: Nice, will rebase the PR tomorrow and address the other stuff as well.

olivergierke:
I'm pretty much done with polish and merge.

xhaggi:
Thanks for taking over.

PR #209, opened 2018-06-14
DATAES-467 - fix sorting by _score if spring-data sort is used

xhaggi:
See https://jira.spring.io/browse/DATAES-467

### DIFF
--- a/src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
@@ -595,6 +595,29 @@ public class ElasticsearchTemplateTests {
 		assertThat(sampleEntities.getContent().get(1).getMessage(), is(sampleEntity2.getMessage()));
 	}
 
+	@Test // DATAES-467
+	public void shouldSortResultsByScore() {
+		// given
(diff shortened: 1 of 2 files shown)


src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
```diff
@@ -595,6 +595,29 @@ public void shouldSortResultsGivenNullLastSortCriteria() {
...
+			.withQuery(matchAllQuery())
+			.withPageable(PageRequest.of(0, 10, Sort.by(Sort.Order.asc("_score"))))
+			.build();
+
+		// when
+		Page<SampleEntity> sampleEntities = elasticsearchTemplate.queryForPage(searchQuery, SampleEntity.class);
+
+		// then
+		assertThat(sampleEntities.getTotalElements(), equalTo(3L));
+	}
```
akonczak: can we check order of elements here not just size of the result

xhaggi: Not really necessary, because without the fix it will throws an exception.

ptahchiev:
I believe this needs to be backported to Kay too.

PR #210, opened 2018-06-16
DATAES-468 Update to 6.3.0

donrw:
Only two test changes.

And, hacked the plugin-descriptor properties.  Anyone know how to get a 6.3.0 lang-expression jar?

- [ X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [X ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAES).
- [X ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [X ] You submit test cases (unit or integration tests) that back your changes.
- [ X] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).


olivergierke:
Given that we don't seem to need any kinds of changes in actual production code, I think we should go ahead with this one for Lovelace. I also wonder if we should introduce a build profile for 6.3 in Kay, so that we can set up a build to verify forward compatibility.

src/test/java/org/springframework/data/elasticsearch/core/SimpleElasticsearchDateMappingTests.java
```diff
@@ -37,6 +39,9 @@
 	@Test
 	public void testCorrectDateMappings() throws NoSuchFieldException, IntrospectionException, IOException {
 		XContentBuilder xContentBuilder = MappingBuilder.buildMapping(SampleDateMappingEntity.class, "mapping", "id", null);
```
olivergierke: Should we rather move to a try-with-resources here?

donrw:
Will fix those tonight.

src/test/java/org/springframework/data/elasticsearch/core/MappingBuilderTests.java
```diff
@@ -53,6 +55,13 @@
 	@Autowired
 	private ElasticsearchTemplate elasticsearchTemplate;
 
+	private String xContentBuilderToString(XContentBuilder builder)
+	{
+		builder.close();
+		ByteArrayOutputStream bos = (ByteArrayOutputStream) builder.getOutputStream();
```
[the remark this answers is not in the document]
donrw: It is possible for other kids of OutputStream to be assigned to the builder. The default is ByteArrayOutputStream.  It is a tricky one since it is a test we control. I was thinking in case the default changed or picked up some configuration that changed it, the problem might be more obvious if it blew up on the cast rather than the String compare.

src/test/java/org/springframework/data/elasticsearch/core/SimpleElasticsearchDateMappingTests.java
donrw: Not sure what the benefit of try-with-resource is in this case.  The close is needed to get the results put on the ByteArrayOutputStream, just like in MappingBuilderTest.  I just didn't pull out the 3 lines liek in MappingBuidlerTest because it was one test.  I debated creating a TestUtil class or the like, so they could share the method, but it didn't seem worth it for two tests.

donrw:
6.3.0 will fail in Kay.  I ran into it specifically in NodeClientFactoryBean.  transport.type local is not supported.  Neither is script.inline.  Also. Setting.builder signature changed.

xhaggi:
>I think we should go ahead with this one for Lovelace.

@olivergierke Sure, shove it in when you're happy with this PR.

>I also wonder if we should introduce a build profile for 6.3 in Kay, so that we can set up a build to verify forward compatibility.

I'm not sure a build profile is enough. How do we handle ES API breaks from 5.x to 6.x?

olivergierke:
> I'm not sure a build profile is enough. How do we handle ES API breaks from 5.x to 6.x?

This upgrade here is from 6.2.2 to 6.3. So I wouldn't expect any breaking changes. I am fine with requiring 6.3 for Lovelace. I just wondered whether on Kay – i.e. Spring Data Elasticsearch 3.0.x – we should add a build profile for 6.3, so that we can run the build with the profile enabled and thus check whether Kay is 6.3 compatible. We do similar things in JPA to make sure we work on different versions of e.g. Hibernate.

donrw:
Changes squashed.

The XContentBuilder interface breaks in a non-backwards compatible way between 6.2.2 and 6.3.0.  But, from other work I am pretty sure that Kay is compatible with 6.2.2. and 6.3.0.  I am not that skilled at manipulating maven, so I have no idea if you can figure out a way to deal with the test in Kay not compiling with 6.3.0 (nor the test in 3.1.X compiling with 6.2.2).

PR #213, opened 2018-06-25
DATAES-290 Make joda-time optional dependency.

ptahchiev:
Make the joda-time optional dependency. Also check if joda-time is in the classpath before registering
the joda-time converters.

This is the first step towards making joda-time as optional dependency. However it is also being used in FacetedPageImpl.java#L161 which
is deprecated and needs to be removed before making the joda-time truly optional.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,7 @@
             <groupId>joda-time</groupId>
             <artifactId>joda-time</artifactId>
             <version>${jodatime}</version>
+            <optional>true</optional>
         </dependency>
 
         <!-- Elasticsearch -->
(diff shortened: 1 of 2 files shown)


xhaggi:
@ptahchiev next time please force-push to your existing branch instead of create new ones. This will safe unnecessary pull requests.

BTW you missed the hyphen `-` between the issue ID and message in your commit message.

olivergierke:
I don't think there's much value in merging this if we don't get the dependency in `FacetedPageImpl` removed. I am suprised to see the cast there. Is ES actually returning JodaTime types there or is this code just broken?

ptahchiev:
I can't say.. i tried to reproduce it to see if elasticsearch really returns joda-time, but i wasn't able to :/

sothawo:
This can be merged if the PR is changed to merge into the branch 4.0.x. In this branch, which uses Elasticsearch 7.3, we do not have the Joda-Time dependency in `FacetedPageImpl` anymore.

sothawo:
took these changes into #308 and merged it into branch 4.0.x

PR #228, opened 2018-11-27
How to configure the number of Thread

Jungle666:
there are 16 threads created，How to configure the number of Thread？
image


mp911de:
This is not a general discussion forum. Please post your question to Stackoverflow.

PR #233, opened 2018-12-11
DATAES-518 - Use scroll for unpaged find operations in ReactiveElasticsearchTemplate.

christophstrobl:


mp911de:
That's merged now.

PR #236, opened 2018-12-21
DATAES-523 - Allow specifying version type.

ivangreene:

Add the ability to specify a version type for documents. This should be mostly passive as the default has been EXTERNAL. The only breaking changes would be to the interfaces `ElasticsearchPersistentEntity` and `ElasticsearchEntityInformation` which expand and thus would break user's custom implementations unless they are extending the provided implementations. Any advice on the preferred way to deal with these additions is appreciated.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
@@ -16,7 +16,6 @@
 package org.springframework.data.elasticsearch.core;
 
 import static org.elasticsearch.client.Requests.refreshRequest;
-import static org.elasticsearch.index.VersionType.EXTERNAL;
 import static org.elasticsearch.index.query.QueryBuilders.moreLikeThisQuery;
 import static org.elasticsearch.index.query.QueryBuilders.wrapperQuery;
 import static org.springframework.data.elasticsearch.core.MappingBuilder.buildMapping;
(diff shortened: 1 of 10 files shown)


src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
```diff
@@ -1175,7 +1177,8 @@ private IndexRequestBuilder prepareIndex(IndexQuery query) {
 			}
 			if (query.getVersion() != null) {
 				indexRequestBuilder.setVersion(query.getVersion());
-				indexRequestBuilder.setVersionType(EXTERNAL);
```
xhaggi: Please remove the static import `org.elasticsearch.index.VersionType.*`

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -1342,7 +1344,8 @@ private IndexRequest prepareIndex(IndexQuery query) {
 			}
 			if (query.getVersion() != null) {
 				indexRequest.version(query.getVersion());
-				indexRequest.versionType(EXTERNAL);
```
xhaggi: Please remove the static import `org.elasticsearch.index.VersionType.EXTERNAL`

ivangreene:
@xhaggi thanks for the response, I've removed the static imports and rebased onto the current master.

xhaggi:
@ivangreene please squash all commits into a single one and we are ready to shove it in.

ivangreene:
@xhaggi done. Thanks

ivangreene:
@xhaggi do you need anything else to merge this? Would like to also backport this to other versions when it makes it into master

xhaggi:
@ivangreene no, thank you.

xhaggi:
@ivangreene By the way, it is not necessary to create backport PR's. I'll take care of it.

PR #237, opened 2019-01-17
DATAES-530 - Add converter based EntityMapper. 

christophstrobl:
The `ElasticsearchEntityMapper` uses SpringData infrastructure instead of Jackson for domain type mapping. This allows to register `Converter` instances to provide more flexibility.

### Cofiguration

```java
@Configuration
public class Config extends AbstractElasticsearchConfiguration {

  @Override
  public RestHighLevelClient elasticsearchClient() {
    return RestClients.create(ClientConfiguration.create("localhost:9200")).rest()
  }

  @Bean
  @Override
  public EntityMapper entityMapper() {

    ElasticsearchEntityMapper entityMapper = new ElasticsearchEntityMapper(elasticsearchMappingContext(),
      new DefaultConversionService());
    entityMapper.setConversions(elasticsearchCustomConversions());

    return entityMapper;
  }
}
```

#### Type Hints

Mapping uses _type hints_ embedded in the document sent to the server to allow generic type mapping. Those type hints are
represented as `_class` attributes within the document on the server and will be written for each aggregate root.

```java
public class Person {

  @Id String id;
  String firstname;
  String lastname;
}
```

```json
{
  "_class" : "com.example.Person",
  "id" : "cb7bef",
  "firstname" : "Sarah",
  "lastname" : "Connor"
}
```

Type hints can be configured to hold custom information. Use the `@TypeAlias` annotation to do so.
Make sure to add types with `@TypeAlias` to the initial entity set (`AbstractElasticsearchConfiguration#getInitialEntitySet`)
to already have entity information available when first reading data from the store.

```java
@TypeAlias("human")
public class Person {

  @Id String id;
  // ...
}
```

```json
{
  "_class" : "human",
  "id" : "..."
}
```

Type hints will not be written for nested Objects unless the properties type is `Object`, an interface or the
actual value type does not match the properties declaration.

#### Geospatial Types

Geospatial types like `Point` & `GeoPoint` are converted into _lat/lon_ pairs.

```java
public class Address {

  String city, street;
  Point location;
}
```
```json
{
  "city" : "Los Angeles",
  "street" : "2800 East Observatory Road",
  "location" : { "lat" : 34.118347, "lon" : -118.3026284 }
}
```

#### Collections

For values inside Collections apply the same mapping rules as for aggregate roots when it comes to _type hints_
and _Custom Conversions_.

```java
public class Person {

  // ...

  List<Person> friends;

}
```
```json
{
  "..." : "...",
  "friends" : [ { "firstname" : "Kyle", "lastname" : "Reese" } ]
}
```

### Maps

The Map key needs to a String to be processed by Elasticsearch.

```java
public class Person {

  // ...

  Map<String, Address> knownLocations;

}
```
```json
{
  "..." : "...",
  "knownLocations" : {
    "arrivedAt" : {
       "city" : "Los Angeles",
       "street" : "2800 East Observatory Road",
       "location" : { "lat" : 34.118347, "lon" : -118.3026284 }
     }
  }
}
```


src/main/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchCustomConversions.java
```diff
@@ -15,22 +15,86 @@
...
 	 *
 	 * @param converters must not be {@literal null}.
 	 */
 	public ElasticsearchCustomConversions(Collection<?> converters) {
-		super(StoreConversions.NONE, converters);
+		super(STORE_CONVERSIONS, converters);
+	}
+
+	@ReadingConverter
+	static class StringToUUIDConverter implements Converter<String, UUID> {
```
mp911de: Let's turn these converters into `enum`s.

src/main/java/org/springframework/data/elasticsearch/core/convert/GeoConverters.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+class GeoConverters {
+
+	static Collection<? extends Object> getConvertersToRegister() {
+		return Arrays.asList(new PointToMapConverter(), new MapToPointConverter(), new GeoPointToMapConverter(),
+				new MapToGeoPointConverter());
+	}
+
+	@WritingConverter
+	static class PointToMapConverter implements Converter<Point, Map> {
```
mp911de: Let's turn these converters into `enum`s.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+	// --> READ
+
+	@Override
+	public <T> T readObject(Map<String, Object> source, Class<T> targetType) {
+		return read(targetType, source);
+	}
+
+	@Override
+	@Nullable
+	public <R> R read(Class<R> type, Map source) {
```
mp911de: Missing generics

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+		return doRead(source, ClassTypeInformation.from((Class<R>) ClassUtils.getUserClass(type)));
+	}
+
+	@Nullable
+	protected <R> R doRead(Map source, TypeInformation<R> typeHint) {
```
mp911de: Missing generics

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+
+		if (source == null) {
+			return null;
+		}
+
+
+
+		if(typeHint.getType().isInterface()) {
```
mp911de: Why is this happening only for interfaces? Wouldn't that be also useful for subtypes?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+					target.put(entry.getKey(), readEntity(targetEntity, (Map) entry.getValue()));
+				}
+
+			}
+		}
+
+		return (R) target;
+	}
+
+	private <R> R readCollectionValue(@Nullable List source, ElasticsearchPersistentProperty property,
```
mp911de: Missing generics and potential NPE in `sourceList.size()`

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+
+		if (ClassTypeInformation.OBJECT.equals(targetType)
+				|| (targetType != null && value.getClass().equals(targetType.getActualType()))) {
+			return value;
+		}
+
+		if (conversionService.canConvert(value.getClass(), targetType.getType())) {
+			return conversionService.convert(value, targetType.getType());
+		}
+
```
mp911de: Do we want to add `Enum` handling here as well?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+		if (value == null) {
+			return null;
+		}
+
+		if (ClassTypeInformation.OBJECT.equals(targetType)
+				|| (targetType != null && value.getClass().equals(targetType.getActualType()))) {
+			return value;
+		}
+
+		if (conversionService.canConvert(value.getClass(), targetType.getType())) {
```
mp911de: Shouldn't that be rather `conversions.hasCustomReadTarget(…)`?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+		}
+
+		Class<?> rawType = targetType.getType();
+		if (conversions.hasCustomReadTarget(source.getClass(), rawType)) {
+			return (R) conversionService.convert(source, rawType);
+		} else if (source instanceof List) {
+			return readCollectionValue((List) source, property, targetType);
+		} else if (source instanceof Map) {
+			return readMapValue((Map<String, Object>) source, property, targetType);
+		} else if (Enum.class.isAssignableFrom(rawType)) {
```
mp911de: The code here is rather source-specific than target-type specific. Do we want to move this into `readSimpleValue(…)`?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+	@Override
+	public Map<String, Object> mapObject(Object source) {
+
+		LinkedHashMap<String, Object> target = new LinkedHashMap<>();
+		write(source, target);
+		return target;
+	}
+
+	@Override
+	public void write(@Nullable Object source, Map sink) {
```
mp911de: Missing generics.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+		List<Object> target = new LinkedList<>();
+		if (!typeHint.getActualType().getType().equals(Object.class)
+				&& conversions.isSimpleType(typeHint.getActualType().getType())) {
+
+			collectionSource.map(this::getWriteSimpleValue).forEach(target::add);
+		} else {
+
+			Streamable.of((Iterable) value).map(it -> {
+
+				if (it == null) {
```
mp911de: Don't we want to retain `null` values in lists?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+		return target;
+	}
+
+	private Object writeCollectionValue(Object value, ElasticsearchPersistentProperty property,
+			TypeInformation<?> typeHint) {
+
+		Streamable collectionSource = value instanceof Iterable ? Streamable.of((Iterable) value)
+				: Streamable.of(ObjectUtils.toObjectArray(value));
+
+		List<Object> target = new LinkedList<>();
```
mp911de: Why do we use a `LinkedList` and not an `ArrayList`?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+		if (!typeHint.getActualType().getType().equals(Object.class)
```
mp911de: How about extracting these checks into a `isSimpleType` method? (See also `writeMapValue(…)`).

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -0,0 +1,676 @@
...
+				continue;
+			}
+
+			Object value = accessor.getProperty(property);
+
+			if (value == null) {
+				continue;
+			}
+
+			if (!conversions.isSimpleType(value.getClass())) {
```
mp911de: We have quite a few `isSimpleType(…)` calls. Do we want to consolidate these into methods?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -878,6 +888,11 @@ public String delete(String indexName, String type, String id) {
 				}
 				return new AggregatedPageImpl<T>(Collections.EMPTY_LIST, response.getScrollId());
 			}
+
```
mp911de: How about introducing an adapter type to not require empty implementations?

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapperTests.java
```diff
@@ -0,0 +1,632 @@
...
+	@Test // DATAES-530
+	public void appliesCustomConverterForRead() {
+		assertThat(entityMapper.read(Inventory.class, shotGunAsMap)).isEqualTo(shotGun);
+	}
+
+	private String pointTemplate(String name, Point point) {
+		return String.format(Locale.ENGLISH, "\"%s\":{\"lat\":%.1f,\"lon\":%.1f}", name, point.getX(), point.getY());
+	}
+
+	private LinkedHashMap<String, Object> writeToMap(Object source) {
```
mp911de: Nit: Simplify to `Map<…>`

PR #245, opened 2019-03-12
DATAES-323 - Use JsonProperty in MappingBuilder.

ivangreene:
Honor the JsonProperty annotation for field names
when building mappings.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.elasticsearch.common.xcontent.XContentBuilder;
 import org.elasticsearch.common.xcontent.XContentType;
 import org.springframework.core.ResolvableType;
@@ -63,6 +64,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * @author Nordine Bittich
  * @author Robert Gruendler
  * @author Petr Kukral
+ * @author Ivan Greene
  */
 class MappingBuilder {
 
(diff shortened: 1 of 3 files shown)


ivangreene:
Point of note: [line 375](https://github.com/spring-projects/spring-data-elasticsearch/blob/436a3213fcd6a8073110b9a2a94784952765ceaf/src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java#L375) in MappingBuilder#isIdField still uses field.getName() because I believe that the persistent property will be using the original field name.

GitProBo:
hi @ivangreene : I found this solution deal with output format, and not useful for ElasticsearchRepository for input fieldname

TestEntity
	@JsonProperty("format_id")
	private Integer formatId;

TestRepository
findAll:   
Test(id=AWnBoHGhRre5ZNxJ_Lxx, formatId=8)

findByFormatId
null

ivangreene:
@GitProBo I'm not entirely sure what you mean. Can you point me to a test or include a snippet where this is an issue? Happy to add a test case

GitProBo:
> @GitProBo I'm not entirely sure what you mean. Can you point me to a test or include a snippet where this is an issue? Happy to add a test case



> @GitProBo I'm not entirely sure what you mean. Can you point me to a test or include a snippet where this is an issue? Happy to add a test case

report updated

ivangreene:
Ok I understand. I will look into this more and try to figure out a solution. Thanks for the comment @GitProBo

roblankey:
And here I thought I was going crazy that this wasn't working any longer. I understand that the autoquerying with the repository potentially falls over as soon as you add JsonProperty renames, and if removing it was explicit in 3.1.0.RELEASE, then it seems like a valid reason.

I wonder if it's worth adding that as another configuration option to enable JsonProperty usage with the caveat that if you, as the dev user, add the JsonProperty attribute and enable the processing of JsonProperty attributes, that you're also going to need to implement your own @Query methods for those fields.

ivangreene:
@roblankey interesting idea for a config option. I myself would opt for that because in my org we are not really using the repositories anyway.
Would like to get some input from a committer on how we should proceed with fixing this.

@xhaggi I see you are assigned on the Jira. Any input on this issue and the type of solution you would like to see? And if not can you tag another committer. Thanks

sothawo:
I don't think that adapting the Mapping in the index is enough. 
When an document is written to the index, the `SpringDataSerializerModifier` from the `SpringDataElasticsearchModule` drops the PropertyWriter for the JsonProperty annotated property because the name of the beanPropertyWriter (that's coming from the annotation) does not match the name of the PersistentProperty.
I'm not sure, if it's better to change this place as well to check for the JsonProperty annotation or to modify the PersitentProperty.

xhaggi:
This will be superseded by #270

ivangreene:
Closing in favor of #270

PR #246, opened 2019-03-14
DATAES-546 - Escaped double quotes in code snippet

g00glen00b:
- [X] You submit test cases (unit or integration tests) that back your changes. (1)
- [X] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only). (2)

This is only a documentation (AsciiDoc) change, so the last three bullet points seem to be irrelevant (unless I missed an author section somewhere in the AsciiDoc). The documentation for the `@Query` annotation does not escape the inner double quotes.


xhaggi:
Thanks, merged.

PR #267, opened 2019-03-24
DATAES-552 - @Query annotation fail when passing over 10 parameters

TaylorOno:
Updated the replacePlaceholders method to use replaceFirst to prevent duplication substitutions

Additional explanations if necessary.

### DIFF
--- a/src/test/java/org/springframework/data/elasticsearch/repository/query/ReactiveElasticsearchStringQueryUnitTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/repository/query/ReactiveElasticsearchStringQueryUnitTests.java
@@ -87,6 +87,18 @@ public class ReactiveElasticsearchStringQueryUnitTests {
 		assertThat(((StringQuery) query).getSource()).isEqualTo(reference.getSource());
 	}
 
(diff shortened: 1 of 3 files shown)


src/test/java/org/springframework/data/elasticsearch/repository/query/ReactiveElasticsearchStringQueryUnitTests.java
```diff
@@ -87,6 +87,18 @@ public void bindsExpressionPropertyCorrectly() throws Exception {
 		assertThat(((StringQuery) query).getSource()).isEqualTo(reference.getSource());
 	}
 
+    @Test // DATAES-552
```
sothawo: the code indentation of the new method is not correct

sothawo:
a solution that I could think of would be to leave the `replace()` call that replaces every instance, but start with the numbering backwards. So you first replace ?11, then ?10, and so on until down to ?1.

I'd suggest to write a test with multiple occurences of the params.

christophstrobl:
Thanks @TaylorOno!
That's been merged and back ported with a slight modification dealing with the issue @sothawo pointed out.

PR #269, opened 2019-04-12
DATAES-561 - Reuse ObjectMapper in ElasticsearchEntityMapper

sothawo:

no new tests needed, unchanged functionality is covered by existing tests.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.elasticsearch.core;
 
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import lombok.RequiredArgsConstructor;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchEntityMapper.java
```diff
@@ -65,6 +66,7 @@
 
 	private final MappingContext<? extends ElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty> mappingContext;
 	private final GenericConversionService conversionService;
+	private final ObjectMapper objectMapper;
```
stefanganzer: Maybe it would be even better to just store references to an `ObjectReader` for HashMap and an `ObjectWriter` obtained from an `ObjectMapper`. They are completly immutable and thread safe, in contrary to the `ObjectMapper`. Even though it is not leaked from the class, and not reconfigured, it would prevent future mishaps.
From https://stackoverflow.com/a/3909846, I understand that's the recommended way of doing it since v2.0.

sothawo: fair point, I changed that. Although the ObjectMapper is not leaked and reconfigured, as you said, it's a little safer now. And we don't have the construction of a reader in every mapToObject call.

PR #270, opened 2019-04-15
DATAES-562 - Custom name attribute for @Field mapping in Elasticsearc…

sothawo:
…hEntityMapper.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.elasticsearch.annotations;
 
+import org.springframework.core.annotation.AliasFor;
+
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Inherited;
@@ -29,12 +31,28 @@ import java.lang.annotation.Target;
  * @author Jonathan Yan
  * @author Jakub Vavrik
  * @author Kevin Leturc
+ * @author Peter-Josef Meisch
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
 @Documented
 @Inherited
 public @interface Field {
+
+	/**
+	 * Alias for {@link #name}.
+	 * @since 3.2
+	 */
+	@AliasFor("name")
+	String value() default "";
+
+	/**
+	 * The <em>name</em> to be used to store the field inside the document.
+	 * <p>If not set, the name of the annotated property is used.
+	 * @since 3.2
+	 */
+	@AliasFor("value")
+	String name() default "";
 
 	FieldType type() default FieldType.Auto;
 
--- a/src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
@@ -18,6 +18,7 @@ package org.springframework.data.elasticsearch.core.mapping;
 import java.util.Arrays;
 import java.util.List;
 
+import org.springframework.data.elasticsearch.annotations.Field;
 import org.springframework.data.elasticsearch.annotations.Parent;
 import org.springframework.data.elasticsearch.annotations.Score;
 import org.springframework.data.mapping.Association;
@@ -26,6 +27,7 @@ import org.springframework.data.mapping.PersistentEntity;
 import org.springframework.data.mapping.model.AnnotationBasedPersistentProperty;
 import org.springframework.data.mapping.model.Property;
 import org.springframework.data.mapping.model.SimpleTypeHolder;
+import org.springframework.util.StringUtils;
 
 /**
  * Elasticsearch specific {@link org.springframework.data.mapping.PersistentProperty} implementation processing
@@ -35,6 +37,7 @@ import org.springframework.data.mapping.model.SimpleTypeHolder;
  * @author Mark Paluch
  * @author Sascha Woo
  * @author Oliver Gierke
+ * @author Peter-Josef Meisch
  */
 public class SimpleElasticsearchPersistentProperty extends
 		AnnotationBasedPersistentProperty<ElasticsearchPersistentProperty> implements ElasticsearchPersistentProperty {
@@ -44,12 +47,14 @@ public class SimpleElasticsearchPersistentProperty extends
 	private final boolean isScore;
 	private final boolean isParent;
 	private final boolean isId;
+	private final String annotatedFieldName;
 
 	public SimpleElasticsearchPersistentProperty(Property property,
 			PersistentEntity<?, ElasticsearchPersistentProperty> owner, SimpleTypeHolder simpleTypeHolder) {
 
 		super(property, owner, simpleTypeHolder);
 
+		this.annotatedFieldName = getAnnotatedFieldName();
 		this.isId = super.isIdProperty() || SUPPORTED_ID_PROPERTY_NAMES.contains(getFieldName());
 		this.isScore = isAnnotationPresent(Score.class);
 		this.isParent = isAnnotationPresent(Parent.class);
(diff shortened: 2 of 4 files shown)


src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -29,13 +31,30 @@
...
+	 *
+	 * @return an empty {@link String} by default.
+	 */
+	@AliasFor("name")
+	String value() default "";
+
+	/**
+	 * The name to be used to store the field inside the document. Alias for {@link #value()}.
+	 *
+	 * @return an empty {@link String} by default.
```
xhaggi: Same here.

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -29,13 +31,30 @@
...
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
 @Documented
 @Inherited
 public @interface Field {
 
+	/**
+	 * The name to be used to store the field inside the document. Alias for {@link #name()}.
+	 *
+	 * @return an empty {@link String} by default.
```
xhaggi: Not sure if we really need this. IMO we should point out that if empty the property name is used.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -44,12 +48,16 @@
 	private final boolean isScore;
 	private final boolean isParent;
 	private final boolean isId;
+	private final boolean isAnnotatedFieldName;
```
xhaggi: I would drop that field and do a null-check in `getFieldName`. `getAnnotatedFieldName` should return `null` in case the name is empty or blank.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -68,13 +76,26 @@ public SimpleElasticsearchPersistentProperty(Property property,
 		}
 	}
 
+	private String getAnnotatedFieldName() {
+
+		if (isAnnotationPresent(Field.class)) {
+
+			final Field annotation = findAnnotation(Field.class);
+			Assert.notNull(annotation, "findAnnotation() returned null on existing Annotation");
```
xhaggi: You don't have to check it twice since `isAnnotationPresent` ensures that the annotation is present.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -68,13 +76,26 @@ public SimpleElasticsearchPersistentProperty(Property property,
...
+	}
+
 	/*
 	 * (non-Javadoc)
 	 * @see org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentProperty#getFieldName()
 	 */
 	@Override
 	public String getFieldName() {
-		return getProperty().getName();
+		return isAnnotatedFieldName ? annotatedFieldName : getProperty().getName();
```
xhaggi:
```java
return annotatedFieldName == null ? getProperty().getName() : annotatedFieldName
```

sothawo: will do when I'm back at home from work ;-)

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
sothawo: the default value must be a constant, so we need an empty String here and cannot use null. To use the alias for the two attributes named _value_ and _name_ was proposed by Mark, this is the same as is used in spring-data-mongodb.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
sothawo: IntelliJ warns me that it may be null, so I had this check (probably using too much Kotlin)

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
sothawo: had it in there in order to not have a hasText() call in every getFieldName call. Switched to null check as you suggested

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
xhaggi: Don't get me wrong, I just want you to adjust the comment to indicate that the field name is optional and by default is set to the property name and not to an empty string.

sothawo: have it adjusted yesterday with the other changes; I can't see the changes on the Github _conversation_ tab for this PR, but on the _Files changed_ tab I can see it.

mp911de: Check out https://github.com/spring-projects/spring-framework/blob/d035593562a1dd39c213143b4c44574051a3fb76/spring-web/src/main/java/org/springframework/web/bind/annotation/RequestParam.java#L66-L77 and https://github.com/spring-projects/spring-framework/blob/44ec6e003762b8bf571b0735746249f4697440e0/spring-web/src/main/java/org/springframework/web/bind/annotation/ResponseStatus.java#L65 for how Spring Framework documents aliased annotation properties.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -44,12 +47,14 @@
 	private final boolean isScore;
 	private final boolean isParent;
 	private final boolean isId;
+	private final String annotatedFieldName;
```
mp911de: In the long run, we might leave caching of model-metadata away at `SimpleElasticsearchPersistentProperty` level and eventually introduce a  `CachingElasticsearchPersistentProperty` (see Mongo's [`CachingMongoPersistentProperty`](https://github.com/spring-projects/spring-data-mongodb/blob/3eb9d790b94e5a40660ef9c5cb262cbe684bbdcb/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/CachingMongoPersistentProperty.java#L29)).

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
xhaggi: @mp911de thanks, that's exactly what I meant, that we don't need the `@return`.
@sothawoIt Sorry I didn't explain it better., would be nice if you could readjust it :)

sothawo: hope I got it right this time.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
sothawo: thanks for the info; I'll have a look at how this is handled in Mongo if I find some time these days

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
xhaggi: Yes you got it ;)

ivangreene:
@xhaggi @sothawo it looks like this may not use the custom field name in MappingBuilder? Or perhaps I'm missing something. But the MappingBuilder.java on master still seems to use `java.lang.reflect.Field#getName` rather than this new method to get either that name or the custom field name

xhaggi:
@ivangreene you're right, I guess we missed that.

sothawo:
@ivangreene, /cc @xhaggi thanks for pointing this out. I will fix this these days ([DATAES-568](https://jira.spring.io/browse/DATAES-568))

PR #271, opened 2019-04-17
DATAES-563 - Repositories configuration assume "elasticsearchTemplate…

sothawo:
…" bean name.

template supports uses name based bean lookup and wants a "elasticsearchTemplate" bean, so this PR adds the bean name to the elasticsearchOperations bean.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/config/AbstractElasticsearchConfiguration.java
+++ b/src/main/java/org/springframework/data/elasticsearch/config/AbstractElasticsearchConfiguration.java
@@ -22,6 +22,7 @@ import org.springframework.data.elasticsearch.core.ElasticsearchRestTemplate;
 
 /**
  * @author Christoph Strobl
+ * @author Peter-Josef Meisch
  * @since 3.2
  * @see ElasticsearchConfigurationSupport
  */
@@ -41,7 +42,7 @@ public abstract class AbstractElasticsearchConfiguration extends ElasticsearchCo
 	 *
 	 * @return never {@literal null}.
 	 */
-	@Bean
+	@Bean(name = {"elasticsearchOperations", "elasticsearchTemplate"})
 	public ElasticsearchOperations elasticsearchOperations() {
 		return new ElasticsearchRestTemplate(elasticsearchClient(), elasticsearchConverter(), resultsMapper());
 	}
--- a/src/test/java/org/springframework/data/elasticsearch/config/ElasticsearchConfigurationSupportUnitTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/config/ElasticsearchConfigurationSupportUnitTests.java
@@ -85,6 +85,20 @@ public class ElasticsearchConfigurationSupportUnitTests {
 		assertThat(context.getBean(ElasticsearchRestTemplate.class)).isNotNull();
 	}
 
+	@Test // DATAES-563
+	public void restConfigContainsElasticsearchOperationsByName() {
+
+		AbstractApplicationContext context = new AnnotationConfigApplicationContext(RestConfig.class);
+		assertThat(context.getBean("elasticsearchOperations")).isNotNull();
+	}
+
+	@Test // DATAES-563
+	public void restConfigContainsElasticsearchTemplateByName() {
+
+		AbstractApplicationContext context = new AnnotationConfigApplicationContext(RestConfig.class);
+		assertThat(context.getBean("elasticsearchTemplate")).isNotNull();
+	}
+
 	@Test // DATAES-504
 	public void reactiveConfigContainsReactiveElasticsearchTemplate() {
 
--- a/src/test/java/org/springframework/data/elasticsearch/config/ElasticsearchConfigurationTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/config/ElasticsearchConfigurationTests.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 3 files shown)


mp911de:
Could you also add a test for repositories using Java-config with `@EnableElasticsearchRepositories`? So far we have only XML-config based integration tests for repositories.

sothawo:
@mp911de do you mean a test like [EnableElasticsearchRepositoriesTests](https://github.com/spring-projects/spring-data-elasticsearch/blob/9818ed0af9e2e9fc3a3ff6deeabb5195c460d53a/src/test/java/org/springframework/data/elasticsearch/config/EnableElasticsearchRepositoriesTests.java#L51) but with a custom `AbstractElasticsearchConfiguration` implementation?

mp911de:
Exactly. Maybe we could even generalize that one to run with different configurations to have a single class with test code and multiple subclasses providing configuration.

sothawo:
added a test for the Java configuration. I have not integrated it with the existing test(s) yet, as I am not sure how yet to proceed with the integration tests when using a rest high level client. The current xml-configured integration tests are using the old node client; the question is how to do integration tests with the rest client. (haven't checked reactive).
Any experiences with [EsIntegTestCase](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/integration-tests.html) or [this maven plugin](https://github.com/alexcojocaru/elasticsearch-maven-plugin)? Is there a better place to discuss this?

src/test/java/org/springframework/data/elasticsearch/config/EnableElasticsearchRepositoriesWithAbstractElasticsearchConfigurationTests.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.data.elasticsearch.repository.config.EnableElasticsearchRepositories;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+/**
+ * @author Peter-Josef Meisch
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration
+public class EnableElasticsearchRepositoriesWithAbstractElasticsearchConfigurationTests {
```
mp911de: `ElasticsearchConfigurationTests` should be a proper name for this one.

mp911de:
Reactive tests use `@ContextConfiguration("classpath:infrastructure.xml")` (see `ReactiveElasticsearchTemplateTests`) to bootstrap a node.

Feel free to overhaul test bootstrapping. We should ideally not depend on a base class, otherwise, this makes it hard to inherit from an abstract test class and the integration test base.

In Spring Data Cassandra, we moved Cassandra initialization to a [test rule](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/util/CassandraRule.java#L58) that can be either consumed through a [base class](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/util/AbstractEmbeddedCassandraIntegrationTest.java#L37) or directly via `@ClassRule`.

For MongoDB and Redis, we spin up servers externally. MongoDB is almost a no-brainer, Redis is spun up in different configurations by a `Makefile`. I'm not sure about a Maven plugin because that raises the question of how to test within the IDE. Java-based servers (Cassandra, Elasticsearch) are ideal candidates to be started once per VM along with the tests to reduce external setup requirements. However, ES likes to break API's and I'm not quite sure how to think about `ESIntegTestCase`.

sothawo:
I renamed the test class.

For testing, can we assume that the test are running in an environment where docker is available? I am thinking about using the [TestContainer Elasticsearch module](https://www.testcontainers.org/modules/elasticsearch/). If this is available in the CI environment, we could make it a prerequisite for local testing. Using a container would allow us to have separate our code from the elasticsearch server and we could use different ES versions for different test. I'd probably also like to use JUnit 5 (keep JUnit 4 as well for the existing tests), to make use of the nested classes for test setups.

I think it'd be best to create an issue for that topic with perhaps some subtasks

mp911de:
TL;DR: I'll go forward and merge your PR. Find answers to your questions below.

> For testing, can we assume that the test are running in an environment where docker is available?

While this works today, I'm a bit torn on Testcontainers. While it's great to get an independent instance of something up and running, it comes with certain costs. Depending on the integration, each test spins up a container and this eats up a lot of bootstrapping time. 

We already see this in Spring Data JDBC and R2DBC where we intentionally reduce the number of integration tests to prevent additional overhead.

From my perspective, I like to have the freedom to do the right thing that depends on a context: By default, you want fast tests. Sometimes you'd like to test against various versions and on rare occasions, you want to spin up a server and run specific tests against that one.

That's how I would prioritize testing. Having a custom rule/extension that takes input from an external configuration (like a system property or property file) would be a decent way to decide what to do (use an embedded server/use Testcontainers/use an external database). That would allow supporting various scenarios and give quite some freedom for test scenarios. This is initially a bit more effort than just plugging a test container instance but in the long run, it reduces the number of problems introduced by yet another technology.

> like to use JUnit 5

Feel free to switch to JUnit 5 leaving JUnit 4 (Vintage) annotations in place if you feel that JUnit solves a substantial issue for you. Currently, we want to leave Spring Data as-is with JUnit 4 with the initial goal to replace Hamcrest with AssertJ before we move forward to JUnit 5.

Back when we promoted Cassandra from a Community module to a Core module we incrementally did cleanups on production code and test code before we could really start adding features. An environment that makes it hard to test things makes it also hard to evolve.

mp911de:
That's merged and polished now. Thanks a lot.

PR #277, opened 2019-04-25
DATAES-569 - Add index operations to reactive client.

christophstrobl:


mp911de:

PR #283, opened 2019-05-26
DATAES-579 - Test code cleanup.

sothawo:
many files changed, but only refactoring:

* replaced usage of JUnit-asserts and hamcrest matchers with assertj.
* refactored the entity classes that were reused in the different tests, so that each test now has the necessary entity classes as nested classes. Changes in an entity now do not affect other tests.
* The refactored nested entity classes now use different names for the index they are stored in to prevent mapping clashes that happened when several tests wrote for example `SampleEntity` mappings in the same index.
* Tests that use repository scanning now are moved to their own packages to prevent repositories from other tests to be scanned on context creation. This has led to problems when these repositories on startup put their mappings in the used index
* removed the fix from [DATAES-576](https://jira.spring.io/browse/DATAES-576) as it is no more necessary.

@mp911de, @christophstrobl , @xhaggi: if you might find the time to have a look at this ;-)


mp911de:
This is a pretty huge PR and a massive test code refactoring that was long overdue. Thanks a lot for that.

When cleansing code, feel free to aggressively drop items that are not needed (e.g. unused properties in test entities). When using Lombok, we should also not implement our own `equals`/`hashCode` methods.

mp911de:

PR #296, opened 2019-07-16
DATAES-603 - Add support for bulk options in Elasticsearch-Template c…

sothawo:
…lasses.

I know that tests are missing, but to test that the options are set into the `BulkRequestBuilder`/`BulkRequest` I would need to somehow intercept the request before it is sent out in the tests. That would mean that I would have to add some listener to the `Elasticsearch(Rest)Template` class just for testing. Doesn't seem good to me. Glad if anybody has an idea how this can be tested.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
@@ -21,12 +21,11 @@ import java.util.stream.Collectors;
 
 import org.elasticsearch.action.update.UpdateResponse;
 import org.elasticsearch.cluster.metadata.AliasMetaData;
-import org.elasticsearch.common.Nullable;
-
 import org.springframework.data.domain.Page;
 import org.springframework.data.elasticsearch.core.convert.ElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentEntity;
 import org.springframework.data.elasticsearch.core.query.AliasQuery;
+import org.springframework.data.elasticsearch.core.query.BulkOptions;
 import org.springframework.data.elasticsearch.core.query.CriteriaQuery;
 import org.springframework.data.elasticsearch.core.query.DeleteQuery;
 import org.springframework.data.elasticsearch.core.query.GetQuery;
@@ -36,6 +35,7 @@ import org.springframework.data.elasticsearch.core.query.SearchQuery;
 import org.springframework.data.elasticsearch.core.query.StringQuery;
 import org.springframework.data.elasticsearch.core.query.UpdateQuery;
 import org.springframework.data.util.CloseableIterator;
+import org.springframework.lang.Nullable;
 
 /**
  * ElasticsearchOperations
@@ -456,18 +456,40 @@ public interface ElasticsearchOperations {
 	UpdateResponse update(UpdateQuery updateQuery);
 
 	/**
-	 * Bulk index all objects. Will do save or update
+	 * Bulk index all objects. Will do save or update.
 	 *
-	 * @param queries
+	 * @param queries the queries to execute in bulk
 	 */
-	void bulkIndex(List<IndexQuery> queries);
+	default void bulkIndex(List<IndexQuery> queries) {
+		bulkIndex(queries, BulkOptions.defaultOptions());
+	}
+
+	/**
+	 * Bulk index all objects. Will do save or update.
+	 *
+	 * @param queries the queries to execute in bulk
+	 * @param bulkOptions options to be added to the bulk request
+	 * @since 3.2
+	 */
+	void bulkIndex(List<IndexQuery> queries, BulkOptions bulkOptions);
 
 	/**
 	 * Bulk update all objects. Will do update
 	 *
-	 * @param queries
+	 * @param queries the queries to execute in bulk
 	 */
-	void bulkUpdate(List<UpdateQuery> queries);
+	default void bulkUpdate(List<UpdateQuery> queries) {
+		bulkUpdate(queries, BulkOptions.defaultOptions());
+	}
+
+	/**
+	 * Bulk update all objects. Will do update
+	 *
+	 * @param queries the queries to execute in bulk
+	 * @param bulkOptions options to be added to the bulk request
+	 * @since 3.2
+	 */
+	void bulkUpdate(List<UpdateQuery> queries, BulkOptions bulkOptions);
 
 	/**
 	 * Delete the one object with provided id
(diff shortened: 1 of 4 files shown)


sothawo:
@mp911de do you think this can still be merged before RC2?

mp911de:
Need to have a look. Likely tomorrow or Monday.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -456,18 +456,38 @@
...
 
 	/**
-	 * Bulk index all objects. Will do save or update
+	 * Bulk index all objects. Will do save or update.
 	 *
-	 * @param queries
+	 * @param queries the queries to execute in bulk
+	 */
+	default void bulkIndex(List<IndexQuery> queries) {
+		this.bulkIndex(queries, null);
```
xhaggi: No need for `this` IMO.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -456,18 +456,38 @@
...
+	void bulkIndex(List<IndexQuery> queries, BulkOptions bulkOptions);
 
 	/**
 	 * Bulk update all objects. Will do update
 	 *
-	 * @param queries
+	 * @param queries the queries to execute in bulk
+	 */
+	default void bulkUpdate(List<UpdateQuery> queries) {
+		this.bulkUpdate(queries, null);
```
xhaggi: Same here.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -764,6 +766,32 @@ public void bulkUpdate(List<UpdateQuery> queries) {
 		}
 	}
 
+	private void setBulkOptions(BulkRequest bulkRequest, BulkOptions bulkOptions) {
+		if (bulkOptions != null) {
```
xhaggi: Should we reduce the scope and do an earlier return if `bulkOptions` is `null`?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
```diff
@@ -635,23 +635,51 @@ private UpdateRequestBuilder prepareUpdate(UpdateQuery query) {
...
 		BulkRequestBuilder bulkRequest = client.prepareBulk();
+		setBulkOptions(bulkRequest, bulkOptions);
 		for (UpdateQuery query : queries) {
 			bulkRequest.add(prepareUpdate(query));
 		}
 		checkForBulkUpdateFailure(bulkRequest.execute().actionGet());
 	}
 
+	private void setBulkOptions(BulkRequestBuilder bulkRequest, BulkOptions bulkOptions) {
+		if (bulkOptions != null) {
```
xhaggi: Same here.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
sothawo: changed it, and added `@Nullable` annotations that I missed before

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplate.java
sothawo: changed as well

xhaggi: Thank you ;)

sothawo:
actually, when having one more look I wonder, if we should expose the ES classes `ActiveShardCount` and `RefreshPolicy` in the `BulkOptions` object or if we shouldn't better use our own objects and map these when setting the values in the request.

mp911de:
Note: RC release date slipped to Aug 1st.

sothawo:
any changes needed on this PR?

src/main/java/org/springframework/data/elasticsearch/core/query/BulkOptions.java
```diff
@@ -0,0 +1,127 @@
...
+import org.elasticsearch.common.unit.TimeValue;
+
+/**
+ * Options that may be passed to an
+ * {@link org.springframework.data.elasticsearch.core.ElasticsearchOperations#bulkIndex(List, BulkOptions)} or
+ * {@link org.springframework.data.elasticsearch.core.ElasticsearchOperations#bulkUpdate(List, BulkOptions)} call.
+ *
+ * @author Peter-Josef Meisch
+ */
+public class BulkOptions {
```
mp911de: With having a Builder in place, what's the reason for non-`final` fields?

xhaggi: +1, pass the builder to a private ctor of `BulkOptions` and set the fields there.

mp911de: It should be rather the other way round, to have a private constructor accepting all `BulkOptions` fields to match Spring Data's style.

sothawo:
* formatting done; changed the `BulkOptions` and the builder; made the parameters not null.
* needed to change two implementations of `ParameterAccessor` classes because of changes in the `org.springframework.data.repository.query.ParameterAccessor` interface.

src/main/java/org/springframework/data/elasticsearch/core/query/BulkOptions.java
```diff
@@ -0,0 +1,113 @@
...
+	public static final class BulkOptionsBuilder {
+		private TimeValue timeout;
+		private WriteRequest.RefreshPolicy refreshPolicy;
+		private ActiveShardCount waitForActiveShards;
+		private String pipeline;
+		private String routingId;
+
+		private BulkOptionsBuilder() {}
+
+		public static BulkOptionsBuilder aBulkOptionsBuilder() {
```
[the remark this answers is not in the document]
sothawo: I supppose you meant to move it to `BulkOptions`? - changed that.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -456,18 +458,40 @@
...
 
 	/**
-	 * Bulk index all objects. Will do save or update
+	 * Bulk index all objects. Will do save or update.
 	 *
-	 * @param queries
+	 * @param queries the queries to execute in bulk
+	 */
+	default void bulkIndex(List<IndexQuery> queries) {
+		bulkIndex(queries, aBulkOptionsBuilder().build());
```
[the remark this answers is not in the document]
sothawo: changed it

src/main/java/org/springframework/data/elasticsearch/core/query/BulkOptions.java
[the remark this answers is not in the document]
mp911de: Sorry, yes, I mixed up the two, it should go to `BulkOptions` as you mentioned.

mp911de:
Thanks a lot. That's merged and polished now.

PR #304, opened 2019-08-16
DATAES-639 - Add ignore_above mapping parameter support.

mars05:

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
@@ -32,6 +32,7 @@ import java.lang.annotation.Target;
  * @author Jakub Vavrik
  * @author Kevin Leturc
  * @author Peter-Josef Meisch
+ * @author Xiao Yu
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
@@ -77,4 +78,9 @@ public @interface Field {
 	boolean includeInParent() default false;
 
 	String[] copyTo() default {};
+
+	/**
+	 * @since 4.0
+	 */
+	int ignoreAbove() default -1;
 }
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
@@ -24,6 +24,7 @@ import java.lang.annotation.Target;
  * @author Artur Konczak
  * @author Mohsin Husen
  * @author Sascha Woo
+ * @author Xiao Yu
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
@@ -48,4 +49,9 @@ public @interface InnerField {
 	String analyzer() default "";
 
 	String normalizer() default "";
+
+	/**
+	 * @since 4.0
+	 */
+	int ignoreAbove() default -1;
 }
--- a/src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
@@ -48,6 +48,7 @@ import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersiste
 import org.springframework.data.mapping.PropertyHandler;
 import org.springframework.data.util.TypeInformation;
 import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 import com.fasterxml.jackson.databind.JsonNode;
@@ -67,6 +68,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * @author Robert Gruendler
  * @author Petr Kukral
  * @author Peter-Josef Meisch
+ * @author Xiao Yu
  */
 class MappingBuilder {
 
@@ -85,6 +87,7 @@ class MappingBuilder {
 	private static final String FIELD_CONTEXT_TYPE = "type";
 	private static final String FIELD_CONTEXT_PRECISION = "precision";
 	private static final String FIELD_DYNAMIC_TEMPLATES = "dynamic_templates";
+	private static final String FIELD_IGNORE_ABOVE = "ignore_above";
 
 	private static final String COMPLETION_PRESERVE_SEPARATORS = "preserve_separators";
 	private static final String COMPLETION_PRESERVE_POSITION_INCREMENTS = "preserve_position_increments";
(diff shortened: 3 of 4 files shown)


src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
```diff
@@ -48,4 +49,6 @@
 	String analyzer() default "";
 
 	String normalizer() default "";
+
+	int ignoreAbove() default -1;
```
sothawo: I'd prefer to use an `Integer`here which as default value is set to `null`.

src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
```diff
@@ -346,6 +348,7 @@ private void addFieldMappingParameters(XContentBuilder builder, Object annotatio
 		String searchAnalyzer = null;
 		String normalizer = null;
 		String[] copyTo = null;
+		int ignoreAbove = -1;
```
sothawo: same as in the annotation: `Integer` initialized to `null`.

src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
```diff
@@ -404,6 +409,9 @@ private void addFieldMappingParameters(XContentBuilder builder, Object annotatio
 		if (copyTo != null && copyTo.length > 0) {
 			builder.field(FIELD_COPY_TO, copyTo);
 		}
+		if (ignoreAbove > -1) {
```
sothawo:
```suggestion
		if (ignoreAbove != null) {
```
after changing to `Integer`

src/test/java/org/springframework/data/elasticsearch/core/MappingBuilderTests.java
```diff
@@ -409,6 +412,41 @@ public void shouldUseFieldNameOnMultiField() throws IOException, JSONException {
...
+		elasticsearchTemplate.createIndex(IgnoreAboveEntity.class);
+		elasticsearchTemplate.putMapping(IgnoreAboveEntity.class);
+
+		// when
+		Map mapping = elasticsearchTemplate.getMapping(IgnoreAboveEntity.class);
+		Map properties = (Map) mapping.get("properties");
+		Map message = (Map) properties.get("message");
+
+		// then
+		java.lang.reflect.Field field = ReflectionUtils.findField(IgnoreAboveEntity.class, "message");
```
sothawo: no need to test if the testclass, which is defined 10 lines later, has this annotation and no need to assert the attribute of the annotation.

This test is about the `MappingBuilder`, so I would not store the mapping in the index, and retrieve it as map and then parse it. This would integration testing. I know, that a lot of tests in this class do this, but the question to answer is _does the MappingBuilder create the correct mapping_? For this you don't need to store the mapping in the index. Have a look at what `shouldBuildMappingsForGeoPoint()` does: Test the JSON string returned from the MappingBuilder to make sure that it contains the correct mapping.

src/test/java/org/springframework/data/elasticsearch/core/MappingBuilderTests.java
```diff
@@ -409,6 +412,41 @@ public void shouldUseFieldNameOnMultiField() throws IOException, JSONException {
...
+
+	/**
+	 * @author Xiao Yu
+	 */
+	@Setter
+	@Getter
+	@NoArgsConstructor
+	@AllArgsConstructor
+	@Builder
+	@Document(indexName = "test-ignore-above-mapping-builder", type = "test", shards = 1, replicas = 0, refreshInterval = "-1")
```
sothawo: `shards = 1`is the default since Elasticsearch 7.0 and Spring Data Elasticsearch 4.0.x

src/test/java/org/springframework/data/elasticsearch/core/MappingBuilderTests.java
```diff
@@ -409,6 +412,41 @@ public void shouldUseFieldNameOnMultiField() throws IOException, JSONException {
...
+	static class IgnoreAboveEntity {
+
+		@Id private String id;
+
+		@Field(type = FieldType.Keyword,ignoreAbove = 10) private String message;
```
sothawo:
```suggestion
		@Field(type = FieldType.Keyword, ignoreAbove = 10) private String message;
```

src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
```diff
@@ -360,6 +363,7 @@ private void addFieldMappingParameters(XContentBuilder builder, Object annotatio
 			searchAnalyzer = fieldAnnotation.searchAnalyzer();
 			normalizer = fieldAnnotation.normalizer();
 			copyTo = fieldAnnotation.copyTo();
+			ignoreAbove = fieldAnnotation.ignoreAbove();
```
sothawo: dd `Assert.isTrue(ignoreAbove > 0, "ignore_above must be a positive value");` to prevent illegal values, same in the `else if` block

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -77,4 +78,6 @@
 	boolean includeInParent() default false;
 
 	String[] copyTo() default {};
+
```
sothawo:
```
/**
 * @since 4.0
 */
```

src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
mars05: But annotation does not support the wrap type.

src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
sothawo: here it can be an `Integer`initialized to null

src/main/java/org/springframework/data/elasticsearch/core/MappingBuilder.java
mars05: Hi, since the annotation defaults to -1, is this assertion necessary?

sothawo: I just checked, 0 is an allowed value, the question is, what do we do if someone configures `-7`? If we want to prevent that, we need the Assertion (but then on `>= 0`), but we could also just leave the Assert out and ignore the negative value. I'd prefer to have the assertion here instead of the user having invalid and ignored code.

mars05:
I have optimized the code, please review it again, thank you.

PR #315, opened 2019-09-09
DATAES-647: Use terms-query instead of multiple should-queries for In and NotIn

rasmusfaber:
This pull request is against 4.0.x as requested (sorry, I don't think I can update the other pull request).


sothawo:
you could have change the target of the PR in GitHub, that's no problem

rasmusfaber:
@sothawo I could easily change the target, but I wanted to change the source as well, and that did not seem to be possible. But anyway: thanks for merging!

PR #317, opened 2019-09-10
DATAES-648 - Unified version compatibility matrix into reference docs.

odrotbohm:
Added information about which Boot version uses which Spring Data Elasticsearch version as well as hints about which versions are still under active maintenance.

Tweaked the readme to rather point to the reference documentation. This needs to be updated to point to the "current" alias once we've reached GA.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -118,16 +118,7 @@ Add the Maven dependency:
 // Always change both files!
 **Compatibility Matrix**
 
-[cols="^,^"]
-|===
-|Spring Data Elasticsearch | Elasticsearch
-|_3.2.x_ (not yet released) |_6.8.1_
-|3.1.x |6.2.2
-|3.0.x |5.5.0
-|2.1.x |2.4.0
-|2.0.x |2.2.0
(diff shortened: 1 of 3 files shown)


carlspring:
Thanks to raising this, @odrotbohm !

I'm linking it to [my question on Stackoverflow](https://stackoverflow.com/questions/57856247/which-versions-of-spring-data-elasticsearch-support-particular-versions-of-elast/).

Similarly, I think a new issue (or maybe even "new issues") should be raised to get the "Hello, World!" examples up-to-date. It should be clearer how to get things working against different versions and these examples should be kept reasonably up-to-date with each release, to save people time from figuring out dependency issues, especially, if they're new to the technology and your project. Currently, the Internet is full of examples that work for different (significanlty) older versions of `spring-boot` and `spring-data-elasticsearch` and `elasticsearch` and it's quite hard to get an accurate point of truth from them, when in fact this source should be your project.

sothawo:
cherry picked commit e72469e into master

odrotbohm:
Thanks, PJ!

PR #324, opened 2019-09-23
DATAES-541 - Removing an alias with ElasticseachRestTemplate does not work properly.

xhaggi:
This fixes DATAES-541 and moves the alias tests from `AliasTests` to `ElasticsearchTemplateTests` so that they are taken into account by the `ElasticsearchRestTemplateTests`. This brought another problem to light because `ElasticsearchRestTemplate#convertAliasResponse` throws a `NoSuchElementException` when the alias response is empty, which is the case if no alias is found for the given index.

@sothawo @mp911de mind taking a look.


src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -1451,11 +1458,11 @@ public void refresh(String indexName) {
 
 		try {
 			JsonNode node = mapper.readTree(aliasResponse);
-
-			Iterator<String> names = node.fieldNames();
-			String name = names.next();
 			node = node.findValue("aliases");
 
+			if (node == null)
```
mp911de: Nit: Wrap `return` with `{…}`.

xhaggi: Done.

PR #327, opened 2019-09-24
DATAES-227 - prepareUpdate does not consider UpdateRequest options.

xhaggi:
This fixes DATAES-227 and besides that considers all possible options of `UpdateRequest` in `prepareUpdate`.

@sothawo mind taking a look.

### DIFF
--- a/src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
@@ -46,6 +46,8 @@ import org.elasticsearch.action.get.MultiGetResponse;
 import org.elasticsearch.action.index.IndexRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.action.support.IndicesOptions;
+import org.elasticsearch.action.update.UpdateRequest;
+import org.elasticsearch.action.update.UpdateRequestBuilder;
 import org.elasticsearch.cluster.metadata.AliasMetaData;
 import org.elasticsearch.index.VersionType;
 import org.elasticsearch.script.Script;
@@ -84,6 +86,7 @@ import org.springframework.data.elasticsearch.core.query.*;
 import org.springframework.data.util.CloseableIterator;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.test.util.ReflectionTestUtils;
 
 /**
  * Base for testing rest/transport templates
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
```diff
@@ -1499,6 +1501,41 @@ public void shouldDoPartialUpdateForExistingDocument() {
...
+				.upsert(doc);
+
+		UpdateQuery updateQuery = new UpdateQueryBuilder() //
+				.withClass(SampleEntity.class) //
+				.withId(documentId) //
+				.withUpdateRequest(updateRequest)
+				.build();
+
+		// when
+		elasticsearchTemplate.update(updateQuery);
```
sothawo: I think, here the created updateQuery should be evaluated and asserted instead of running the the update againast ES and then checking results. These are unit tests and not integration tests.

Besides that, although Spring provides the `ReflectionTestUtils` class I don't like this approach to use reflection to make parts of a tested class accessible against the visibility rules.

xhaggi: Sure, but most tests use the same approach.

xhaggi: >I think, here the created updateQuery should be evaluated and asserted instead of running the the update againast ES and then checking results. These are unit tests and not integration tests.

After a second look it makes no sense to assert the created `UpdateQuery` because the issue is fixed in `ElasticsearchTemplate#prepareUpdate` and not in `UpdateQueryBuilder` or `UpdateQuery`. If we do not want to include ES in this test, we need to access the private methods `ElasticsearchTemplate#prepareUpdate` and `ElasticsearchRestTemplate#prepareUpdate`. Let me know your preference.

>Besides that, although Spring provides the ReflectionTestUtils class I don't like this approach to use reflection to make parts of a tested class accessible against the visibility rules.

@mp911de What are the best practices here?

mp911de: We go on a case-by-case basis. We try to test using unit tests as much as possible as integration tests add up to buildtime. Using reflection is sometimes the only way to access things especially when working with 3rd-party API.

sothawo: my first priority would be to make this a unit test and do not use ES. I know we have lots of these unittests that use a running Es, but we should get rid of these as much as possible. They have been a pita in the last time, with the conflicting indices and entities. This has become a little better with my last big test refactoring, but we should avoid it.

So use the RefelctionTestUtils to access the private methods to test, this I think is better than for example changing the visibility of these methods to make them testable.

Might be worth a rewrite to move the prepareXXX methods out of the Template classes, maybe we should add an issue to the [rewrite issues](https://jira.spring.io/issues/?filter=16441);

xhaggi: Sounds good to me. Ill let you know when it's done.

xhaggi: Done. Mind taking another look?

PR #329, opened 2019-10-03
DATAES-654 - Add Junit5 support.

sothawo:
This PR adds the setup for using JUnit 5 tests (the existing JUnit4 based test work as before).

First, normal unit tests that don't need an Elasticsearch cluster can be written straight away.
For tests that need an Elasticsearch cluster and the integration with Spring, there is a new `@SpringDataElasticsearchTest` annotation which combines the new `SpringDataElasticsearchExtension` and the `SpringExtension`. 

The `SpringDataElasticsearchExtension` starts an Elasticsearch node on initialization of the first test class that uses this annotation. This cluster is used for all tests until the last one has finished - it's not recreated for every test class. This should make the tests run faster when more and more tests are switched to use this.

The sample tests (`JUnit5SampleRestClientBasedTests` for example) show how to set up a test class that has a correctly configured application context.

In addition to that it is possible by setting the environment variable SPRING_DATA_ELASTICSEARCH_TEST_CLUSTER_URL to not start a cluster but use an existing externally started instance:

```
SPRING_DATA_ELASTICSEARCH_TEST_CLUSTER_URL=http://my-elastic-host:9200 mvn -Dtest=JUnit5SampleRestClientBasedTests test
```


pom.xml
```diff
@@ -242,7 +243,26 @@
 			</exclusions>
 		</dependency>
 
-		<dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
```
mp911de: We need to consider JUnit 5 support from a build-perspective. We need to evaluate what's the best approach, ideally, we can import JUnit 5 dependencies in the same way as we do it with Junit 4.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
```diff
@@ -0,0 +1,83 @@
...
+	private static final Lock initLock = new ReentrantLock();
+
+	@Override
+	public void beforeAll(ExtensionContext extensionContext) throws Exception {
+		LOGGER.debug("initializing");
+
+		ClusterConnection clusterConnection = getClusterConnection(extensionContext);
+		Class<?> testClass = extensionContext.getRequiredTestClass();
+
+		AnnotationSupport.findAnnotatedFields(testClass, SpringDataElasticsearch.class).forEach(field -> {
```
mp911de: Probably we should use only the built-in injection mechanism instead of assigning values to `static` fields.

pom.xml
sothawo: that's right, but none of the parent spring-data poms provides a junit 5 version. I would have preferred to not define the dependency and the version here

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
sothawo: This runs before the spring context is set up. When Spring is set up, this information must already be available to be able to inject it into the spring beans

mp911de: Yeah, therefore it would make sense to associate the connection details with the Spring context. I'm not sure how much customization we can apply for `SpringExtension`. I think we would need to call `ConfigurableListableBeanFactory.registerResolvableDependency(…)` at some point. Maybe @sbrannen can hint us in the right direction.

pom.xml
mp911de: Can you file a ticket against https://github.com/spring-projects/spring-data-build/?

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
sothawo: I was thinking about extending `SpringExtension` but am not sure if this would be the right way. So I'd be glad for an hint how to get this logic of setting up the ClusterConnection only once for the test run and shutitng it down at the end of the tests coupled to Spring injection to have this info as a bean available.

mp911de: Basically, we need to bridge resources provided by our Spring Data ES Extension into the Spring context so Spring beans can utilize it. We're on the same page that the connection details lifecycle is managed by our Spring Data ES Extension.

pom.xml
sbrannen: You also do not want to manage the individual versions for JUnit 5 artifacts.

Rather, you should import the `junit-bom` for that.

And if you want to use JUnit Jupiter, you'll probably want to depend on the `junit-jupiter` aggregating artifact (similar to a Spring Boot starter POM).

src/test/java/org/springframework/data/elasticsearch/JUnit5SampleRestClientBasedTests.java
```diff
@@ -0,0 +1,65 @@
...
+ * class demonstrating the setup of a JUnit 5 test in Spring Data Elasticsearch that uses the rest client. The
+ * ContextConfiguration must include the test class itself (and provide a
+ * {@link org.springframework.data.elasticsearch.junit.jupiter.ClusterConnectionInfo} bean) and the
+ * {@link ElasticsearchRestTemplateConfiguration} class
+ *
+ * @author Peter-Josef Meisch
+ */
+@SpringDataElasticsearchTest
+@ContextConfiguration(
+		classes = { JUnit5SampleRestClientBasedTests.class, ElasticsearchRestTemplateConfiguration.class })
```
sbrannen: It's not a good idea to use a test class as a pseudo `@Configuration` class (even it "lite bean mode").

Rather, you should use a static nested `@Configuration` class within the test class for such purposes, and that class can then `@Import` additional configuration like the `ElasticsearchRestTemplateConfiguration`.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ClusterConnectionInfo.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * the information that is injected into the test classes by the {@link SpringDataElasticsearchExtension}.
+ *
+ * @author Peter-Josef Meisch
+ */
+public final class ClusterConnectionInfo {
+	private boolean useSsl = false;
+	private String host;
+	private int port;
+	private Client client = null;
```
sbrannen: I'd make all the fields final.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ElasticsearchRestTemplateConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+ * Configuration for Spring Data Elasticsearch using
+ * {@link org.springframework.data.elasticsearch.core.ElasticsearchRestTemplate}. The required
+ * {@link ClusterConnectionInfo} bean must be provided by the testclass.
+ *
+ * @author Peter-Josef Meisch
+ */
+@Configuration
+public class ElasticsearchRestTemplateConfiguration extends AbstractElasticsearchConfiguration {
+
+	@Autowired private ClusterConnectionInfo clusterConnectionInfo;
```
sbrannen: I wouldn't make this a required bean. If it's not present in the `ApplicationContext`, can't you just create a default instance for the user?

That would also remove the requirement that the user always configure a `@Bean` method for that if the defaults are sufficient.

src/test/java/org/springframework/data/elasticsearch/JUnit5SampleTransportClientBasedTests.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+@SpringDataElasticsearchTest
+@ContextConfiguration(
+		classes = { JUnit5SampleTransportClientBasedTests.class, ElasticsearchTemplateConfiguration.class })
+@DisplayName(" a sample JUnit 5 test with transport client")
+public class JUnit5SampleTransportClientBasedTests {
+
+	// Test environment setup
+	// this will be set by the SpringDataElasticsearchExtension, must be static
+	@SpringDataElasticsearch static private ClusterConnectionInfo clusterConnectionInfo;
```
sbrannen: Why does this have to be static?

If you absolutely need it at the class level, have you considered using `@TestInstance(PER_CLASS)` semantics?

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
```diff
@@ -0,0 +1,83 @@
...
+ * shutdown at the end of all tests.
+ *
+ * @author Peter-Josef Meisch
+ */
+public class SpringDataElasticsearchExtension implements BeforeAllCallback {
+
+	public static final String SPRING_DATA_ELASTICSEARCH_TEST_CLUSTER_URL = "SPRING_DATA_ELASTICSEARCH_TEST_CLUSTER_URL";
+	private static final Logger LOGGER = LoggerFactory.getLogger(SpringDataElasticsearchExtension.class);
+
+	private static final String STORE_KEY = ClusterConnection.class.getSimpleName();
```
sbrannen: I wouldn't use the simple name. Somebody else might have a class named that.

Why not just use the `Class` reference?

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
sbrannen: > Yeah, therefore it would make sense to associate the connection details with the Spring context. I'm not sure how much customization we can apply for `SpringExtension`. I think we would need to call `ConfigurableListableBeanFactory.registerResolvableDependency(…)` at some point. Maybe @sbrannen can hint us in the right direction.

If that's what you want to do, you could implement that in a `TestExecutionListener` (Spring API, not JUnit Platform) like I demonstrated here: https://github.com/spring-projects/spring-framework/issues/12947.

sbrannen: You might also want to consider implementing a custom `org.springframework.test.context.ContextCustomizer`.

Spring Boot's testing support relies heavily on the `ContextCustomizer` SPI for many of its features.

mp911de: Thanks a lot, Sam. These are the APIs we were looking for.

sothawo: thanks for all this information; I'll have a deeper look a t this the next week

mp911de:
Related: spring-projects/spring-data-build#906

mp911de:
Spring Data Build includes now `junit-jupiter` and the vintage engine. We can remove dependency-related changes from `pom.xml` in this PR.

pom.xml
sothawo: fixed that (thanks @mp911de for adding that to spring-data-build)

sothawo:
I adaapted the pom; will look after the other points when the other PR regarding the Mappers is done, that one is more important

pom.xml
mp911de: Spring Data Build is pulling in [`junit-jupiter` and `junit-vintage-engine` dependencies](https://github.com/spring-projects/spring-data-build/blob/master/parent/pom.xml#L945-L955).

sothawo:
My problem is, that when implementing a `TestExecutionListener` or a `ContextCustomizer` I do not have access to the `ClusterConnection` object that is stored in the JUnit5 ExtensionContext. `TestExecutionListener` gets a Spring `TestContext` which is independent from the JUnit5 `ExtensionContext`.

I though about extending `SpringExtension` and putting the `ClusterConnection` into the `TestContext`, but the method `getTestContextManage()` is private and so I have no access from a derived class.

sothawo:
I now implemented this using a `ThreadLocal` setting it up in our extension and then retrieving it in the configuration class.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchTest.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Combines the {@link SpringDataElasticsearchExtension} and the {@link SpringExtension}.
+ *
+ * @author Peter-Josef Meisch
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@ExtendWith(SpringDataElasticsearchExtension.class)
+@ExtendWith(SpringExtension.class)
+public @interface SpringDataElasticsearchTest {
```
mp911de: I'm not sure whether we need this, but how about renaming this annotation to `@IntegrationTest`? It probably makes sense to remove `@Extend(SpringExtension)` here and introduce it in Spring-specific tests. The reason is that `SpringDataElasticsearchExtension` should handle primarily the Elasticsearch connectivity aspect. The integration with Spring is a secondary concern. See also comments on `SpringDataElasticsearchExtension`.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * This extension class check in the {@link #beforeAll(ExtensionContext)} call if there is already a Elasticsearch
+ * cluster connection defined in the root store. If no, the connection to the cluster is defined according to the
+ * configuration, starting a local node if necessary. The connection is stored and will be closed when the store is
+ * shutdown at the end of all tests.
+ *
+ * @author Peter-Josef Meisch
+ */
+public class SpringDataElasticsearchExtension implements BeforeAllCallback {
```
mp911de: It would make sense to implement also `ParameterResolver` here. This is, to give non-Spring tests the chance to connect to Elasticsearch using a client so tests can inject `ClusterConnection` into the test method or a setup method.

Furthermore, we need to address the integration with Spring. Given the ordering aspect, we might want to make this class also implement `ContextCustomizerFactory` to customize the Spring context before it bootstraps (requires `META-INF/spring.factories`). We could register a resolvable dependency with 

```java
	@Override
	public ContextCustomizer createContextCustomizer(Class<?> testClass, List<ContextConfigurationAttributes> configAttributes) {
		return this::customizeContext;
	}

	private void customizeContext(ConfigurableApplicationContext context, MergedContextConfiguration mergedConfig) {

		context.getBeanFactory()
				.registerResolvableDependency(ClusterConnectionInfo.class, ClusterConnection
						.getClusterConnectionInfo());
	}
```

so that configuration classes can declare a `@Autowired ClusterConnectionInfo` and so we connect both pieces. 

I wasn't able to achieve the same using a `TestExecutionListener` as the listener invocation seems too late for our purposes.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ElasticsearchTemplateConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+ * Configuration for Spring Data Elasticsearch using
+ * {@link org.springframework.data.elasticsearch.core.ElasticsearchTemplate}.
+ *
+ * @author Peter-Josef Meisch
+ */
+@Configuration
+public class ElasticsearchTemplateConfiguration extends ElasticsearchConfigurationSupport {
+
+	@Bean
+	public Client elasticsearchClient() {
```
mp911de: We can inject `ClusterConnectionInfo` as a method parameter. Alternatively, a `@Autowired ClusterConnectionInfo` field.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ClusterConnectionInfo.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+
+/**
+ * the information about the ClusterConnection.
+ *
+ * @author Peter-Josef Meisch
+ */
+public final class ClusterConnectionInfo {
+	final private boolean useSsl;
```
mp911de: Nit: `private final` instead of `final private`.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ClusterConnectionInfo.java
```diff
@@ -0,0 +1,94 @@
...
+	final private String host;
+	final private int port;
```
mp911de: What type of port is this? HTTP or Transport Client?

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ClusterConnection.java
```diff
@@ -0,0 +1,113 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * This class manages the connection to an Elasticsearch Cluster, starting a local one if necessary. The information
+ * about the ClusterConnection is stored in a static ThreadLocal<ClusterConnectionInfo> and can be accessed with the
+ * {@link ClusterConnection#getClusterConnectionInfo()} method.
+ *
+ * @author Peter-Josef Meisch
+ */
+class ClusterConnection implements ExtensionContext.Store.CloseableResource {
```
mp911de: It would make sense to store both information, the endpoint and the handle to a cluster runner (Node client, Test containers) in the Jupiter context for separation of concerns.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchTest.java
sothawo: The reason I created this annotation was to combine the two extensions in the right order, so that `SpringDataElasticsearchExtension` runs before `SpringExtension. One can still use the extensions on their own wit `@ExtendWith` like I did in the new `JUnit5ClusterConnectionTests`test

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
sothawo: implemented this; I just didn't see the last connecting step from my implementation yesterday and then context customizing ;-) Now Autowiring works and parameter resolution as well

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ClusterConnection.java
sothawo: The ClusterConnection is stored in the ExtensionContext. Or do you mean having separate objects for the different connection types? I think, we will remove the client part when removing transport client stuff, so we then will only need the host/port/ssl part. Even when adding support for TestContainers (optional test depenedency?), the connection to the ES in a TestContainer will be defined by the URL

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchTest.java
mp911de: Then let's shoot for `@SpringIntegrationTest` and `@IntegrationTest` annotations for easier consumption in the tests.

sothawo: done that.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/IntegrationTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+/**
+ * Wraps the {@link SpringExtension}.
```
mp911de: Nit: `SpringDataElasticsearchExtension`

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	/*
+	 * (non javadoc)
+	 * no need to check the paramaterContext and extensionContext here, this was done before in supportsParameter.
+	 */
+	@Override
+	public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext)
+			throws ParameterResolutionException {
+		return ClusterConnection.getClusterConnectionInfo();
```
mp911de: Storing `getClusterConnectionInfo` within the `ExtensionContext.store` would remove the `ThreadLocal` dependency from the lookup.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ClusterConnection.java
mp911de: See comment below in the actual extension.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/ClusterConnectionInfo.java
```diff
@@ -0,0 +1,94 @@
...
+import org.elasticsearch.client.Client;
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+
+/**
+ * the information about the ClusterConnection.
+ *
+ * @author Peter-Josef Meisch
+ */
+public final class ClusterConnectionInfo {
```
mp911de: Nit: Let's add a bit of proper Javadoc to this class and the builder.

src/test/java/org/springframework/data/elasticsearch/junit/jupiter/SpringDataElasticsearchExtension.java
```diff
@@ -0,0 +1,105 @@
...
+
+	private static final String STORE_KEY = ClusterConnection.class.getSimpleName();
+
+	private static final Lock initLock = new ReentrantLock();
+
+	@Override
+	public void beforeAll(ExtensionContext extensionContext) throws Exception {
+		initLock.lock();
+		try {
+			ExtensionContext rootContext = extensionContext.getRoot();
```
mp911de: In this method we can guarantee single-threaded behavior, therefore I'd suggest looking up `ClusterConnection`  and storing it in the `ExtensionContext.store`, too. With migrating `resolveParameter(…)` to use the extension store, we no longer depend on threading for the parameter lookup. The only integration that needs to worry about `Thread` affinity is how we pass `ClusterConnection` into the Spring context.

PR #340, opened 2019-11-02
Dataes 631   consolidate query objects

sothawo:
Draft PR.


src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -339,34 +330,20 @@
 	 * @param clazz
 	 * @return
 	 */
-	<T> long count(CriteriaQuery query, Class<T> clazz);
+	<T> long count(Query query, @Nullable Class<T> clazz);
```
mp911de: `Query` currently carries types and indices. Typically, we keep these table-like coordinates on method level so a `Query` object can be applied to various (tables/collections/)indices. I wonder whether it would make sense to extract indices/types into a type that holds both information. We could then add these coordinates to `count(…)` via `count(Query, Class<?>, Index...)`. That way, we separate the actual criteria from execution-specific targets.

Ideally, we can expose `count` with the following overloads:

* `long count(Query query, Class<?> entityClass);`
* `long count(Query query, Index... indexes);`
* `long count(Query query, @Nullable Class<?> entityClass, Index... indexes);`

We should also drop the generic type `T` as there's no outbound usage of `T`.

src/main/java/org/springframework/data/elasticsearch/core/AbstractQueryMapper.java
```diff
@@ -0,0 +1,120 @@
...
+import org.springframework.data.elasticsearch.core.query.CriteriaQuery;
+import org.springframework.data.elasticsearch.core.query.Query;
+import org.springframework.data.elasticsearch.core.query.SearchQuery;
+import org.springframework.lang.Nullable;
+
+/**
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public abstract class AbstractQueryMapper<R> {
```
mp911de: We probably don't want a generic type as the query mapper should always return a concrete type that can be used with the store. There's probably not much value in sharing an `AbstractQueryMapper` across the Rest and the Transport Client implementation.

sothawo:
@mp911de: next iteration:

I moved the `IndexCoordinates` from being an inner class in `EntityOperations` out and changed it so it also can handle multiple index names and type names.

The `count()` methods from `ElasticsearchOperations` now have additionally the possibility to use `IndexCoordinates` instead of implicitly having the index and type extracted from the entity or from the `Query`.

There now is a `SearchRequestFactory` which is used - at the moment only - for the `count()` implementations. This factory takes the index info from the passed-in `IndexCoordinates`, if these are not set, from the entity. It does **not** use the index/type values from the `Query` object, index names and types will be removed from `Query`once all the queries are rewritten to use the SearchRequestFactory` and the `IndexCoordinates`. 

What I have not started to implement is the QueryMapper, part of this mapping is already contained for example in the `prepareSort` method of the `SearchRequestFactory`, for this I still need to pass the Entity class down into this method. That will be changed later.

`SearchQuery`interface is removed as well as the not needed `EsClient` interface.

Am I going in the right direction?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -326,36 +326,34 @@
...
-	<T> long count(CriteriaQuery query, Class<T> clazz);
-
-	/**
-	 * return number of elements found by given query
-	 *
-	 * @param query
-	 * @return
-	 */
-	<T> long count(CriteriaQuery query);
+	default long count(Query query, @Nullable Class<?> clazz) {
```
mp911de: Ideally, we can make the entity class non-nullable (if we manage getting rid of the index coordinates inside of `Query`).

src/main/java/org/springframework/data/elasticsearch/core/IndexCoordinates.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.springframework.data.elasticsearch.core;
+
+import org.springframework.util.Assert;
+
+/**
+ * Value object encapsulating index name(s) and index type(s).
+ *
+ * @author Mark Paluch
+ * @author Christoph Strobl  @author Peter-Josef Meisch
```
mp911de: Nit: Line break, `@since`

src/main/java/org/springframework/data/elasticsearch/core/IndexCoordinates.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class IndexCoordinates {
+
+	private final String[] indexNames;
+	private final String[] typeNames;
+
+	public IndexCoordinates(String[] indexNames) {
```
mp911de: Let's make constructors `private` so creation works only using static factory methods and mutators (`of(…)`, `withIndex(…)`, `withType(…)`).

src/main/java/org/springframework/data/elasticsearch/core/SearchRequestFactory.java
```diff
@@ -0,0 +1,326 @@
...
+
+/**
+ * Factory class to create Elasticsearch {@link org.elasticsearch.action.search.SearchRequest} and
+ * {@link org.elasticsearch.action.search.SearchRequestBuilder}} instances from
+ * {@link org.springframework.data.elasticsearch.core.query.Query} objects.
+ *
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public class SearchRequestFactory {
```
mp911de: Let's keep that type package-protected.

src/main/java/org/springframework/data/elasticsearch/core/SearchRequestFactory.java
```diff
@@ -0,0 +1,326 @@
...
+				sort.missing("_first");
+			} else if (order.getNullHandling() == Sort.NullHandling.NULLS_LAST) {
+				sort.missing("_last");
+			}
+
+			return sort;
+		}
+	}
+
+	private Pair<QueryBuilder, QueryBuilder> getQueryAndFilter(Query query) {
```
mp911de: Let's resolve the need for a `Pair` by splitting that method into two methods.

src/main/java/org/springframework/data/elasticsearch/core/SearchRequestFactory.java
```diff
@@ -0,0 +1,326 @@
...
+			elasticsearchFilter = new CriteriaFilterProcessor().createFilterFromCriteria(criteriaQuery.getCriteria());
+		} else {
+			throw new IllegalArgumentException("unhandled Query implementation " + query.getClass().getName());
+		}
+
+		return new Pair<>(elasticsearchQuery, elasticsearchFilter);
+	}
+
+	@Nullable
+	private String[] retrieveIndexNameFromPersistentEntity(@Nullable Class clazz) {
```
mp911de: The class should be non-nullable and we should use `getRequiredPersistentEntity(…)` to avoid NPEs.

This method is called when we do not have index coord specified, so having a persistent entity is a requirement. Same for `retrieveTypeFromPersistentEntity`.

sothawo:
@mp911de 

I'm making progress. The work is not complete. What I did up to now:
- refactor creation of different requests/requestbuilders to `RequestFactory`
- only use the `Query` class in `ElasticsearchOperations` instead of a special method for every subclass (not complete yet)
- introduce `IndexCoordinates` to get away from index name(s)/type(s) being passed in the `Query`objects - types are deprecated in ES anyway
- I nearly removed all elasticsearch classes from the `ElasticsearchOperations` API (not complete yet)
- I removed all the Facet related classes that were deprecated since 2016 (DATAES-180)
- consolidated logic from `ElasticsearchTemplate` and `ElasticsearchRestTemplate` into `AbstractElasticsearchTemplate` and into default methods of `ElasticsearchOperations`

I'm not yet content with the many calls of `IndexCoordinates.of()` in the tests, need some better approach there.

If you find the time please have a look.
Do you know why the travis build for this PR fails?

mp911de:
The default Travis environment does not provide Java 8 anymore. Only the older Travis environments come with Java 8. 

We're currently in the process of enabling Spring Data builds (the build itself, using Spring Data on Java9+ works for quite some time) on Java 11. There's a [branch](https://github.com/spring-projects/spring-data-elasticsearch/tree/issue/DATAES-690) and [build](https://travis-ci.org/spring-projects/spring-data-elasticsearch/builds/615162499) that work on Java 11 with `mvn verify`.

sothawo:
I think it's time to merge this

src/main/java/org/springframework/data/elasticsearch/core/IndexCoordinates.java
```diff
@@ -0,0 +1,72 @@
...
+		Assert.notNull(indexNames,"indexNames must not be null");
+		return new IndexCoordinates(indexNames);
+	}
+
+	private IndexCoordinates(String[] indexNames) {
+		Assert.notEmpty(indexNames, "indexNames may not be null or empty");
+		this.indexNames = indexNames;
+	}
+
+	public IndexCoordinates withTypes(String... typeNames) {
```
mp911de: Let's follow the immutable approach and create a new instance of `IndexCoordinates` here so `IndexCoordinates` can be used as a fully immutable object.

src/main/java/org/springframework/data/elasticsearch/core/IndexCoordinates.java
```diff
@@ -0,0 +1,72 @@
...
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public class IndexCoordinates {
+
+	private final String[] indexNames;
+	private String[] typeNames = new String[0];
+
+	public static IndexCoordinates of(String... indexNames) {
+		Assert.notNull(indexNames,"indexNames must not be null");
```
mp911de: Let's also ensure that `indexNames` contains at least one element so we can get rid of nullability in `getIndexName()`

src/main/java/org/springframework/data/elasticsearch/core/IndexCoordinates.java
```diff
@@ -0,0 +1,72 @@
...
+
+	public String getIndexName() {
+		return indexNames != null && indexNames.length > 0 ? indexNames[0] : null;
+	}
+
+	public String[] getIndexNames() {
+		return indexNames;
+	}
+
+	public String getTypeName() {
```
mp911de: Nit: `@Nullable`

src/main/java/org/springframework/data/elasticsearch/core/IndexCoordinates.java
```diff
@@ -0,0 +1,72 @@
...
+		Assert.notNull(typeNames, "typeNames must not be null");
+		this.typeNames = typeNames;
+		return this;
+	}
+
+	public String getIndexName() {
+		return indexNames != null && indexNames.length > 0 ? indexNames[0] : null;
+	}
+
+	public String[] getIndexNames() {
```
mp911de: Not sure, we might want to return a copy of the array to prevent mutation (e.g `getIndexNames()[0] = "foo"`)

PR #341, opened 2019-11-03
DATAES-680 - ReactiveElasticsearchTemplate-should-use-the-count-API.

sothawo:
Currently the reactive stack in order to do a count does a search and counts the returned flux. This means that all records are read to get the count which is quite not performant.

This PR uses the _count_ API of Elasticsearch to do a count.

This is somehow related to #340 but I want to keep this in a separate PR, as this feature should be merged into 3.2.x regardless of the rewriting in the current master branch.


sothawo:
I agree. But doing this as long as the TransportClient is supported means additional work which will be obsolete in the near future.

PR #352, opened 2019-12-03
DATAES-634 - Rearrange methods in Template API.

sothawo:
I added the `IndexOperations`, `DocumentOperations`,` SearchOperations` for non-reactive and `ReactiveDocumentOperations`, `ReactiveSearchOperations`for the reactive part. It's just a pure refactoring, extracting the methods to the corresponding interfaces, leaving some helper methods in the `(Reactive)ElasticSearchOperations` and pushing some default methods down into the implementations.

In `AbstractElasticsearchTemplate` and `ReactiveElasticsearchTemplate` I heavily reordered the methods in the code, because I grouped them in `// region` and `// endregion` comments to have them displayed by group in IntelliJ 
grafik


mp911de:
That's starting to look pretty good. Let's move index operations to its own implementation class and add deprecated `default` methods that forward to `indexOps().…` methods (see [`DefaultedRedisConnection`](https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/connection/DefaultedRedisConnection.java#L54-L63). That way, we preserve backward compatibility and at the same time, we help users to migrate to the new API.

Design-wise, nested operations objects (see MongoDB `BulkOperations` and [`DefaultBulkOperations`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/DefaultBulkOperations.java#L86)) typically hold a reference to the actual template API without subclassing `Abstract…Template`.

I wonder whether we should leave document and search operations in the `…ElasticsearchOperations` interface. I found it quite confusing why `exists(…)` is on a document interface while `count(…)` is on the search operations.

sothawo:
`count()` and `exists()` are in different REST API groups when looking at Elasticsearch, so I mirrored this here, found it strange as well. I guess this is because count needs to do a search internally while exists has a key to access.

What are the naming conventions for interfaces/implementations? Up to now `*Operations` are interfaces and `*Template` are the implementations. In Mongo you use `Operations`for both?

mp911de:
Implementation is named `…Template` and the interface is `…Operations` for top-level entry-points. For some reason for nested operations, we follow the `Default…Operations` naming scheme where implementations are typically package-private which is different from the top-level scheme.

sothawo:
So for the `IndexOperations`I would create a `DefaultTransportIndexOperations`and a `DefaultRestIndexOperations` - Although I'd prefer to keep _Rest_ out of the name and just name it `DefaultIndexOperations`for the RestClient version.

mp911de:
`DefaultIndexOperations` sounds just about right for the REST implementation. Since these types are not part of the public API, names shouldn't hurt anyone besides us.

sothawo:
> That's starting to look pretty good. Let's move index operations to its own implementation class and add deprecated `default` methods that forward to `indexOps().…` methods (see [`DefaultedRedisConnection`](https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/connection/DefaultedRedisConnection.java#L54-L63). That way, we preserve backward compatibility and at the same time, we help users to migrate to the new API.

If I make the `DefaultIndexOperations`package private and delegate the index methods to this from the existing `ElasticsearchRestTemplate`, - when deprecating them, we need a `ElasticsearchOperations.getIndexOperations()` so the user can get the current `IndexOperations`implementation, right?

mp911de:
Exactly.

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	<T> T get(GetQuery query, Class<T> clazz, IndexCoordinates index);
+
+	/**
+	 * Execute a multiGet against elasticsearch for the given ids
+	 *
+	 * @param query
+	 * @param clazz
+	 * @param index
+	 * @return
```
xhaggi: nit: missing `@param` and `@return` docs?

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	default void bulkIndex(List<IndexQuery> queries, IndexCoordinates index) {
+		bulkIndex(queries, BulkOptions.defaultOptions(), index);
+	}
+
+	/**
+	 * Bulk index all objects. Will do save or update.
+	 *
+	 * @param queries the queries to execute in bulk
+	 * @param bulkOptions options to be added to the bulk request
```
xhaggi: nit: missing `@param index` docs

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -0,0 +1,135 @@
...
+	 * @param clazz
+	 * @param index
+	 * @return
+	 */
+	<T> List<T> multiGet(Query query, Class<T> clazz, IndexCoordinates index);
+
+	/**
+	 * Bulk index all objects. Will do save or update.
+	 *
+	 * @param queries the queries to execute in bulk
```
xhaggi: nit: missing `@param index` docs

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public interface DocumentOperations {
+
+	/**
+	 * Index an object. Will do save or update
+	 *
+	 * @param query
```
xhaggi: nit: missing `@param index` docs

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -0,0 +1,135 @@
...
+	 * @param queries the queries to execute in bulk
+	 * @param bulkOptions options to be added to the bulk request
+	 * @since 3.2
+	 */
+	void bulkIndex(List<IndexQuery> queries, BulkOptions bulkOptions, IndexCoordinates index);
+
+	/**
+	 * Bulk update all objects. Will do update
+	 *
+	 * @param queries the queries to execute in bulk
```
xhaggi: nit: missing `@param index` docs

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	default void bulkUpdate(List<UpdateQuery> queries, IndexCoordinates index) {
+		bulkUpdate(queries, BulkOptions.defaultOptions(), index);
+	}
+
+	/**
+	 * Bulk update all objects. Will do update
+	 *
+	 * @param queries the queries to execute in bulk
+	 * @param bulkOptions options to be added to the bulk request
```
xhaggi: I am too lazy to mark all places where the Javadoc misses some `@param`.

sothawo:
the missing javadoc params wer already in there, I just moved the methods (although this is no excuse not to add them)

xhaggi:
> the missing javadoc params wer already in there, I just moved the methods (although this is no excuse not to add them)

Ok.

sothawo:
I split out the `IndexOperations` implementations for Transport/Rest, the reactive part has no IndexOperations yet, so nothing to do here.

For splitting out the implementations of `DocumentOperations`and `SearchOperations`I'd like to create separate Issues. to not have this PR hanging around too long

sothawo:
I have refactored the code into the different implementations of the `*Operations` interfaces. `ElasticsearchTemplate` and `ElasticsearchRestTemplate` now are only delegating to the respective `*Operations` implementations with deprecated methods, they don't even keep a reference to the (transport/rest) client any longer.

Java doc of the interface classes is updated, there are 2 methods (alias and suggests) where we still have Elasticsearch classes in the API, that's for later cleanup.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -126,450 +80,275 @@ public ElasticsearchRestTemplate(RestHighLevelClient client, ElasticsearchConver
...
 	@Override
-	public boolean addAlias(AliasQuery query, IndexCoordinates index) {
-		IndicesAliasesRequest request = requestFactory.indicesAddAliasesRequest(query, index);
-		try {
-			return client.indices().updateAliases(request, RequestOptions.DEFAULT).isAcknowledged();
-		} catch (IOException e) {
-			throw new ElasticsearchException("failed to update aliases with request: " + request, e);
-		}
+	@Deprecated
+	public boolean createIndex(String indexName, Object settings) {
```
mp911de: Deprecations should happen on the interface. The implementation is typically not visible to users as we recommend using the `ElasticsearchOperations` interface directly.

src/main/java/org/springframework/data/elasticsearch/core/AbstractDefaultDocumentOperations.java
```diff
@@ -0,0 +1,83 @@
...
+	@Override
+	public void delete(Query query, Class<?> clazz, IndexCoordinates index) {
+		Assert.notNull(query, "Query must not be null.");
+		SearchRequest searchRequest = requestFactory.searchRequest(query, clazz, index);
+		DeleteQuery deleteQuery = new DeleteQuery();
+		deleteQuery.setQuery(searchRequest.source().query());
+		delete(deleteQuery, index);
+	}
+
+	// region Helper methods
```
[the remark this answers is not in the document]
sothawo: I agree that this is more readable, the `// region` comments allow IntelliJ to fold/unfold these regions in the editor and the structure view; that format is recognized by VSCode as well according to Jetbrains docs

src/main/java/org/springframework/data/elasticsearch/core/BaseElasticsearchOperations.java
```diff
@@ -0,0 +1,49 @@
...
+
+	protected final ElasticsearchConverter elasticsearchConverter;
+	protected final RequestFactory requestFactory;
+
+	public BaseElasticsearchOperations(ElasticsearchConverter elasticsearchConverter) {
+		this.elasticsearchConverter = elasticsearchConverter;
+		requestFactory = new RequestFactory(elasticsearchConverter);
+	}
+
+	public ElasticsearchPersistentEntity<?> getPersistentEntityFor(Class<?> clazz) {
```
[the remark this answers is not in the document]
sothawo: I renamed it here to `getRequiredPersistentEntity`that just forwards

src/main/java/org/springframework/data/elasticsearch/core/BaseElasticsearchOperations.java
```diff
@@ -0,0 +1,49 @@
...
+		Assert.isTrue(clazz.isAnnotationPresent(Document.class), "Unable to identify index name. " + clazz.getSimpleName()
```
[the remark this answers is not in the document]
sothawo: I rmeoved the check here. At the moment ther is much information related to the creation of an index in the `@Document` annotation, so until we have that split from the repository handling, we'll need it.

src/main/java/org/springframework/data/elasticsearch/core/DefaultDocumentOperations.java
```diff
@@ -0,0 +1,155 @@
...
+
+	@Override
+	public String index(IndexQuery query, IndexCoordinates index) {
+		IndexRequest request = requestFactory.indexRequest(query, index);
+		try {
+			String documentId = client.index(request, RequestOptions.DEFAULT).getId();
+
+			// We should call this because we are not going through a mapper.
+			if (query.getObject() != null) {
+				setPersistentEntityId(query.getObject(), documentId);
```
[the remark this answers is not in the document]
sothawo: I agree with that

src/main/java/org/springframework/data/elasticsearch/core/BaseElasticsearchOperations.java
```diff
@@ -0,0 +1,49 @@
...
+				+ " is not a Document. Make sure the document class is annotated with @Document(indexName=\"foo\")");
+		return elasticsearchConverter.getMappingContext().getRequiredPersistentEntity(clazz);
+	}
+
+	public IndexCoordinates getIndexCoordinatesFor(Class<?> clazz) {
+		ElasticsearchPersistentEntity entity = getPersistentEntityFor(clazz);
+		return IndexCoordinates.of(entity.getIndexName()).withTypes(entity.getIndexType());
```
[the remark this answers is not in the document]
sothawo: done.

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -0,0 +1,137 @@
...
+	default void bulkIndex(List<IndexQuery> queries, IndexCoordinates index) {
+		bulkIndex(queries, BulkOptions.defaultOptions(), index);
+	}
+
+	/**
+	 * Bulk index all objects. Will do save or update.
+	 *
+	 * @param queries the queries to execute in bulk
+	 * @param bulkOptions options to be added to the bulk request
+	 * @since 3.2
```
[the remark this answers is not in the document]
sothawo: a copy-leftover from the old class

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -46,451 +31,11 @@
...
-	 * @param query
-	 * @param clazz
-	 * @param index
-	 * @param <T>
-	 * @return
-	 */
-	<T> Page<T> moreLikeThis(MoreLikeThisQuery query, Class<T> clazz, IndexCoordinates index);
-
-	ElasticsearchPersistentEntity getPersistentEntityFor(Class clazz);
+public interface ElasticsearchOperations extends IndexOperations, DocumentOperations, SearchOperations {
```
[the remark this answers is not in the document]
sothawo: ok, fixed that

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
sothawo: done

sothawo:
I still have `ReactiveElasticsearchOperations extends ReactiveDocumentOperations, ReactiveSearchOperations `- still need to change that according to the non-reactive part so that we have no inheritance but delegation and deprecation

mp911de:
Seems we had a misunderstanding regarding `DocumentOperations` and `SearchOperations`. We should deprecate only `IndexOperation` methods on `ElasticsearchOperations` and add `@Deprecated` forwarding methods for index operations while leaving document- and search operations unchanged. Therefore, Reactive operations are not affected at all.

`@Deprecated` forwarding methods on the `ElasticsearchOperations` interface should look like:

```
	@Deprecated
	default boolean indexExists(Class<?> clazz) {
		return getIndexOperations().indexExists(clazz);
	}
```

sothawo:
Do we then need the `SearchOperations` and `DocumentOperations` interfaces extracted from the `ElasticsearchOperations`? and their default implementations?

mp911de:
It makes sense to keep the interfaces. I'm not sure about the default implementations – maybe, if they make things easier.

sothawo:
So `ElasticsearchOperations` will implement `SearchOperations` and `DocumentOperations` and in the `ElasticsearchOperations`implementations either implement these or delegate to the internally available `DocumentOperations`and `SearchOperations` without deprecating these calls (then I think it makes sense to remerge these implementations back to the Operations implementation).

And the methods of the `IndexOperations` which is not implemented are kept as deprecated in the `ElasticsearchOperations`interface and implemented with calls to the getIndexOperations() implementation

sothawo:
index operations are separated out, document and search as interfaces with the implementations like before in `Elasticsearch(Rest)Template`

PR #355, opened 2019-12-10
DATAES-239 - Query for null values.

sothawo:
A `null` argument value in a query now leads to the construction of a 
```
must_not {
  exists {
    "field": "fieldname"
  }
}
```
subquery, allowing to specify queries searching for documents where a given field does **not** exist.


xhaggi:
@sothawo personally, I don't like to mix code changes and formatting. This makes a review somewhat difficult and also hides the code changes in that commit.

sothawo:
you're right about the formatting, but here it mostly came from applying the autoformatting which I normally do on every file I work upon

sidian123:
I think it's better to also support query keyword  `IS_NULL` in query method. 

For example:

```java
long findAllByNameIsNull();
```

PR #357, opened 2019-12-12
DATAES-709 - Add parameter to include defaults on get settings

xhaggi:
https://jira.spring.io/browse/DATAES-709

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractDefaultIndexOperations.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractDefaultIndexOperations.java
@@ -20,7 +20,9 @@ import static org.springframework.util.StringUtils.*;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.elasticsearch.action.admin.indices.settings.get.GetSettingsResponse;
 import org.elasticsearch.common.collect.MapBuilder;
+import org.elasticsearch.common.settings.Settings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.data.elasticsearch.ElasticsearchException;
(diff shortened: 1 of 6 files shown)


xhaggi:
@sothawo IMO the method name should be `getSettings()`. Should we mark the current one as deprecated, add the new one (separate commit) and after that I add the overloads?

sothawo:
I just had a look at the Elasticsearch docs. The `RestHighLevelClient` can execute a `GetSettingsRequest`. So we should use that instead of doing the request with the low level client. I assume that the existing code here comes from the times when the high level client did not yet support this request (as far as I remember that was in ES 5). 
This would then enable us for both cases (`DefaultIndexOperations`and `DefaultTransportIndexOperations`) to use the same code to parse the response and move that to `AbstractDefaultIndexOperations`.

As for the naming: I think we should add the different `getIndexSettings()` methods to the `IndexOperations` interface and document them with `since 4.0`. The existing should then be changed to deprecated with a default implementation that uses the new one:
```
	/**
	 * Get settings for a given indexName.
	 *
	 * @param indexName the name of the index
	 * @return the settings
	 * @deprecated since 4.0, use getSettings(indexName) instead.
	 */
         @Deprecated
	default Map<String, Object> getSetting(String indexName) {
		return getSettings(indexName);
	}
```
No need for a different issue for that.

xhaggi:
@sothawo Thanks for pointing out that the HighLevelClient now supports it. I have addressed your comments and force-pushed the changes.

sothawo:
As the `getIndexSettings()`methods are part of the `IndexOperations`interface I would not add _Index_ in the method name. I think, `IndexOperations.getSettings(...)` is enough. 

Having looked at that again, I think we can even remove the `getSetting(...)`methods from `IndexOperations`, as this interface was just introduced in the master branch; no need to deprecate something that was never released.

But we would need to change the deprecated definitions of the old methods in `ElasticsearchOperations`:

```
	/**
	 * Get settings for a given indexName.
	 *
	 * @param indexName the name of the index
	 * @return the settings
	 * @deprecated since 4.0, use {@link IndexOperations#getSettings(String)} )} instead}
	 */
	@Deprecated
	default Map<String, Object> getSetting(String indexName) {
		return getIndexOperations().getSettings(indexName);
	}
```

Sorry, hadn't noticed that before.

xhaggi:
@sothawo all done and force-pushed.

PR #358, opened 2019-12-13
DATAES-252 - fix Offset calculation logic

ghiron:
Currently startRecord is calculated by multiplying page number to page size. Should use PageRequest.getOffset method

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: java
 
+dist: trusty
+
 jdk:
   - oraclejdk8
 
--- a/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
@@ -143,6 +143,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * @author Peter-Josef Meisch
  * @author Mathias Teier
  * @author Gyula Attila Csorogi
+ * @author Guillaume Hiron
  */
 public class ElasticsearchRestTemplate extends AbstractElasticsearchTemplate
 		implements ElasticsearchOperations, EsClient<RestHighLevelClient>, ApplicationContextAware {
@@ -1338,7 +1339,7 @@ public class ElasticsearchRestTemplate extends AbstractElasticsearchTemplate
 		}
 
 		if (query.getPageable().isPaged()) {
-			startRecord = query.getPageable().getPageNumber() * query.getPageable().getPageSize();
+			startRecord = query.getPageable().getOffset();
 			sourceBuilder.size(query.getPageable().getPageSize());
 		}
 		sourceBuilder.from(startRecord);


sothawo:
Thanks you for this PR. This behaviour however was just recently addressed in #353 on the master branch (DATAES-187). DATAES-252 was a duplicate of DATAES-187 that wasn't noticed. I just marked and closed it as duplicate.
As this is no bug in the 3.2 release, we did not backport the change but only implement it in the master branch.
So sorry, that we won't merge this, but we really appreciate that you contribute to this project.

ghiron:
Hi,
Thanks for reply! 
I saw it has been fixed on master branch but I don't get why this is not a bug in 3.2 release? My code is based on 3.2 release. I use an Offset based PageRequest to query ES but my code fails because this piece of code.

sothawo:
What `PageRequest`do you use? The implementations I find in spring-data-commons and that are used in Spring Data Elasticsearch are all set up with a page number and a size and have the following implementations:
```

    public int getPageSize() {
        return this.size;
    }

    public int getPageNumber() {
        return this.page;
    }

    public long getOffset() {
        return (long)this.page * (long)this.size;
    }
```
So for these `PageRequest` implementations it makes no difference if the `getOffset()` is used or the product from the other two values.

ghiron:
I use this home made page request : 
```
public class OffsetPageRequest implements Pageable {

    private final int offset;
    private final Sort sort;
    private final int limit;

    public OffsetPageRequest(int limit) {
        this(0, limit, Sort.unsorted());
    }


    public OffsetPageRequest(int offset, int limit, Sort sort) {
        this.limit = limit;
        this.offset = offset;
        this.sort = sort;
    }

    @Override
    public int getPageNumber() {
        return 0;
    }

    @Override
    public int getPageSize() {
        return limit;
    }

    @Override
    public long getOffset() {
        return offset;
    }

    @Override
    public Sort getSort() {
        return sort;
    }

    @Override
    public Pageable next() {
        return null;
    }

    @Override
    public Pageable previousOrFirst() {
        return null;
    }

    @Override
    public Pageable first() {
        return null;
    }

    @Override
    public boolean hasPrevious() {
        return false;
    }
}

```
This page request allows me to use offset base paginatinon.

sothawo:
But this implementation violates the contract defined by the Pageable interface:

```
	/**
	 * Returns the page to be returned.
	 *
	 * @return the page to be returned.
	 */
	int getPageNumber();

	/**
	 * Returns the number of items to be returned.
	 *
	 * @return the number of items of that page
	 */
	int getPageSize();

	/**
	 * Returns the offset to be taken according to the underlying page and page size.
	 *
	 * @return the offset to be taken
	 */
	long getOffset();
```
note especially
> according to the underlying page and page size

So a `Pageable` implementations must provide these values, the concept of a page in the first place has a number and a size, the offset value being provided is just an additional value.

So it's not a bug to not use the offset but the page number and size.

ghiron:
I understand your point but IMHO Pagination and offset are not related :  
* In standard implementation `offset` starts at 0 and is incremented by `page size` at each page. 
* But how can I request page 10, size 20 but with an inital offset of 2 for eg ( records 202 -> 242) ?

I can make this simple class(which does not violates contract I think) : 
```
  public class PageRequestWithOffset extends PageRequest {
        public long getOffset() {
            return super.getOffset() + 2;
        }
    }
```

I have exactly the same issue that the one described on https://jira.spring.io/browse/DATAES-187 . I cannot explain better than this ticket you mentionned before.

The way it is implemented in spring-data-elasticsearch overrides the logic we can add in `PageRequest`. 

This override blocks flexibility (and blocks me) : It is not possible to use paging offset other than multiples of page size. 
No offense (really) but I don't understand why this limitation is valid enough to fix master but not backport it on 3.2.

sothawo:
It's implemented in master as a new feature/improvement. In released branches only bugfixes are added for Service Releases. This is no bug, as the paging behaviour like it was released with 3.2 is working as designed.

ghiron:
Ok. I guess I have to wait next release !
Thank you for your time and discussion

PR #359, opened 2019-12-13
DATAES-672 - Introduce SearchHit and SearchHits types to enrich search results

sothawo:
This Draft-PR  introduces `SearchHit<T>` which will be returned from the `SearchOperations` methods instead of a plain `T`. It currently has as additional information the score of the hit, more stuff like highlights, sort etc. will be added later with new issues for that.

Repositories don't return these as it would be impossible to distinguish at runtime of a repository method should return `List<T>` or `List<SearchHit<T>>`. This will be adresses in a separate issue to enable this for repositories by introducing `SearchHits<T>`.

This is a draft PR as these changes are in the non-reactive part currently. I will add the `SearchHit` to the reactive part in the next days probably.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
@@ -14,13 +14,12 @@ import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.index.query.MoreLikeThisQueryBuilder;
-import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.search.suggest.SuggestBuilder;
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
-import org.springframework.data.domain.Page;
 import org.springframework.data.elasticsearch.ElasticsearchException;
+import org.springframework.data.elasticsearch.core.aggregation.AggregatedPage;
 import org.springframework.data.elasticsearch.core.convert.ElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.convert.MappingElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.document.SearchDocumentResponse;
@@ -98,26 +97,29 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	public <T> CloseableIterator<T> stream(Query query, Class<T> clazz, IndexCoordinates index) {
 		long scrollTimeInMillis = TimeValue.timeValueMinutes(1).millis();
 		return StreamQueries.streamResults(startScroll(scrollTimeInMillis, query, clazz, index),
-				scrollId -> continueScroll(scrollId, scrollTimeInMillis, clazz), this::clearScroll);
+				scrollId -> continueScroll(scrollId, scrollTimeInMillis, clazz), this::searchScrollClear);
 	}
 
 	@Override
-	public <T> Page<T> moreLikeThis(MoreLikeThisQuery query, Class<T> clazz, IndexCoordinates index) {
+	public <T> CloseableIterator<SearchHit<T>> searchForStream(Query query, Class<T> clazz, IndexCoordinates index) {
+		long scrollTimeInMillis = TimeValue.timeValueMinutes(1).millis();
+		return StreamQueries.streamResults(searchScrollStart(scrollTimeInMillis, query, clazz, index),
+				scrollId -> searchScrollContinue(scrollId, scrollTimeInMillis, clazz), this::searchScrollClear);
+	}
+
+	@Override
+	public <T> AggregatedPage<SearchHit<T>> search(MoreLikeThisQuery query, Class<T> clazz, IndexCoordinates index) {
 
 		Assert.notNull(query.getId(), "No document id defined for MoreLikeThisQuery");
 
 		MoreLikeThisQueryBuilder moreLikeThisQueryBuilder = requestFactory.moreLikeThisQueryBuilder(query, index);
 
-		return queryForPage(new NativeSearchQueryBuilder().withQuery(moreLikeThisQueryBuilder).build(), clazz, index);
+		return searchForPage(new NativeSearchQueryBuilder().withQuery(moreLikeThisQueryBuilder).build(), clazz, index);
 	}
 
 	@Override
-	public <T> List<T> queryForList(Query query, Class<T> clazz, IndexCoordinates index) {
-		return queryForPage(query, clazz, index).getContent();
-	}
-
-	@Override
-	public <T> List<Page<T>> queryForPage(List<? extends Query> queries, Class<T> clazz, IndexCoordinates index) {
+	public <T> List<AggregatedPage<SearchHit<T>>> multiSearchForPage(List<? extends Query> queries, Class<T> clazz,
+			IndexCoordinates index) {
 		MultiSearchRequest request = new MultiSearchRequest();
 		for (Query query : queries) {
 			request.add(requestFactory.searchRequest(query, clazz, index));
@@ -126,7 +128,8 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	}
 
 	@Override
-	public List<Page<?>> queryForPage(List<? extends Query> queries, List<Class<?>> classes, IndexCoordinates index) {
+	public List<AggregatedPage<? extends SearchHit<?>>> multiSearchForPage(List<? extends Query> queries,
+			List<Class<?>> classes, IndexCoordinates index) {
 		MultiSearchRequest request = new MultiSearchRequest();
 		Iterator<Class<?>> it = classes.iterator();
 		for (Query query : queries) {
@@ -135,9 +138,10 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 		return doMultiSearch(queries, classes, request);
 	}
 
-	private <T> List<Page<T>> doMultiSearch(List<? extends Query> queries, Class<T> clazz, MultiSearchRequest request) {
+	private <T> List<AggregatedPage<SearchHit<T>>> doMultiSearch(List<? extends Query> queries, Class<T> clazz,
+			MultiSearchRequest request) {
 		MultiSearchResponse.Item[] items = getMultiSearchResult(request);
-		List<Page<T>> res = new ArrayList<>(queries.size());
+		List<AggregatedPage<SearchHit<T>>> res = new ArrayList<>(queries.size());
 		int c = 0;
 		for (Query query : queries) {
 			res.add(elasticsearchConverter.mapResults(SearchDocumentResponse.from(items[c++].getResponse()), clazz,
@@ -146,10 +150,10 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 		return res;
 	}
 
-	private List<Page<?>> doMultiSearch(List<? extends Query> queries, List<Class<?>> classes,
-			MultiSearchRequest request) {
+	private List<AggregatedPage<? extends SearchHit<?>>> doMultiSearch(List<? extends Query> queries,
+			List<Class<?>> classes, MultiSearchRequest request) {
 		MultiSearchResponse.Item[] items = getMultiSearchResult(request);
-		List<Page<?>> res = new ArrayList<>(queries.size());
+		List<AggregatedPage<? extends SearchHit<?>>> res = new ArrayList<>(queries.size());
 		int c = 0;
 		Iterator<Class<?>> it = classes.iterator();
 		for (Query query : queries) {
@@ -160,17 +164,6 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	}
 
 	abstract protected MultiSearchResponse.Item[] getMultiSearchResult(MultiSearchRequest request);
-
-	protected List<String> extractIds(SearchResponse response) {
-		List<String> ids = new ArrayList<>();
-		for (SearchHit hit : response.getHits()) {
-			if (hit != null) {
-				ids.add(hit.getId());
-			}
-		}
-		return ids;
-	}
-
 	// endregion
 
 	// region Helper methods
@@ -195,7 +188,7 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	public abstract SearchResponse suggest(SuggestBuilder suggestion, IndexCoordinates index);
 
 	/**
-	 * @param clazz
+	 * @param clazz the entity class
 	 * @return the IndexCoordinates defined on the entity.
 	 * @since 4.0
 	 */
--- a/src/main/java/org/springframework/data/elasticsearch/core/SearchHit.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/SearchHit.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * Encapsulates the found data with additional information from the search.
+ * 
+ * @param <T> the result data class.
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public class SearchHit<T> {
+
+	private final String id;
+	private final float score;
+	private final T content;
+
+	public SearchHit(@Nullable String id, float score, T content) {
+		this.id = id;
+		this.score = score;
+		this.content = content;
+	}
+
+	@Nullable
+	public String getId() {
+		return id;
+	}
+
+	/**
+	 * @return the score for the hit.
+	 */
+	public float getScore() {
+		return score;
+	}
+
+	/**
+	 * @return the object data from the search.
+	 */
+	public T getContent() {
+		return content;
+	}
+
+	@Override
+	public String toString() {
+		return "SearchHit{" +
+				"id='" + id + '\'' +
+				", score=" + score +
+				", content=" + content +
+				'}';
+	}
+}
--- a/src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import reactor.core.publisher.Flux;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.springframework.data.elasticsearch.core.aggregation.AggregatedPage;
+import org.springframework.data.elasticsearch.core.aggregation.impl.AggregatedPageImpl;
+
+/**
+ * Utility class with helper methods for working with {@link SearchHit}.
+ * 
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public final class SearchHitSupport {
+
+	private SearchHitSupport() {}
+
+	/**
+	 * unwraps the data contained in a SearchHit for different types containing SearchHits if possible
+	 * 
+	 * @param result the object, list, page or whatever containing SearchHit objects
+	 * @return a corresponding object where the SearchHits are replaced by their content if possible, otherwise the
+	 *         original object
+	 */
+	public static Object unwrapSearchHits(Object result) {
+
+		if (result == null) {
+			return result;
+		}
+
+		if (result instanceof SearchHit<?>) {
+			return ((SearchHit<?>) result).getContent();
+		}
+
+		if (result instanceof List<?>) {
+			return ((List<?>) result).stream() //
+					.map(SearchHitSupport::unwrapSearchHits) //
+					.collect(Collectors.toList());
+		}
+
+		if (result instanceof AggregatedPage<?>) {
+			AggregatedPage<?> page = (AggregatedPage<?>) result;
+			List<?> list = page.getContent().stream().map(o -> unwrapSearchHits(o)).collect(Collectors.toList());
+			return new AggregatedPageImpl<>(list, null, page.getTotalElements(), page.getAggregations(), page.getScrollId(),
+					page.getMaxScore());
+
+		}
+
+		if (result instanceof Stream<?>) {
+			return ((Stream<?>) result).map(SearchHitSupport::unwrapSearchHits);
+		}
+
+		if (result instanceof SearchHits<?>) {
+			SearchHits<?> searchHits = (SearchHits<?>) result;
+			return unwrapSearchHits(searchHits.getSearchHits());
+		}
+
+		if (result instanceof Flux) {
+			Flux<?> flux = (Flux<?>) result;
+			return flux.map(SearchHitSupport::unwrapSearchHits);
+		}
+
+		return result;
+	}
+}
--- a/src/main/java/org/springframework/data/elasticsearch/core/SearchHits.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/SearchHits.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.springframework.data.util.Streamable;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Encapsulates a list of {@link SearchHit}s with additional information from the search.
+ *
+ * @param <T> the result data class.
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public class SearchHits<T> implements Streamable<SearchHit<T>> {
+
+	private final List<? extends SearchHit<T>> searchHits;
+	private final long totalHits;
+	private final float maxScore;
+
+	/**
+	 * @param searchHits must not be {@literal null}
+	 * @param totalHits
+	 * @param maxScore
+	 */
+	public SearchHits(List<? extends SearchHit<T>> searchHits, long totalHits, float maxScore) {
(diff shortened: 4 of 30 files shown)


mp911de:
Spring Data MongoDB has a bit of code to distinguish between `List<T>`, `List<GeoResult<T>>` and `GeoResults<T>`, see [`AbstractMongoQuery`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/AbstractMongoQuery.java#L133-L142) and [`MongoQueryMethod`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/MongoQueryMethod.java#L197-L212).

src/main/java/org/springframework/data/elasticsearch/core/SearchHit.java
```diff
@@ -0,0 +1,76 @@
...
+ * @param <T> the result data class.
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public final class SearchHit<T> {
+
+	private final float score;
+	private final T hit;
+
+	public static <T> Builder<T> builder() {
```
mp911de: Let's avoid the builder pattern for simple objects such as this one. Having a public constructor is fine for a simple domain object. All of our [domain objects](https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/geo/GeoPage.java) in commons follow this idea which makes it easy to use and reuse such types.

src/main/java/org/springframework/data/elasticsearch/core/SearchHit.java
```diff
@@ -0,0 +1,76 @@
...
+
+	private SearchHit(float score, T hit) {
+		this.score = score;
+		this.hit = hit;
+	}
+
+	/**
+	 * @return the score for the hit.
+	 */
+	float getScore() {
```
mp911de: Nit: `public`

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -95,14 +95,15 @@ public void delete(Query query, Class<?> clazz, IndexCoordinates index) {
...
-	public <T> CloseableIterator<T> stream(Query query, Class<T> clazz, IndexCoordinates index) {
+	public <T> CloseableIterator<SearchHit<T>> stream(Query query, Class<T> clazz, IndexCoordinates index) {
 		long scrollTimeInMillis = TimeValue.timeValueMinutes(1).millis();
 		return StreamQueries.streamResults(startScroll(scrollTimeInMillis, query, clazz, index),
 				scrollId -> continueScroll(scrollId, scrollTimeInMillis, clazz), this::clearScroll);
 	}
 
 	@Override
-	public <T> Page<T> moreLikeThis(MoreLikeThisQuery query, Class<T> clazz, IndexCoordinates index) {
+	public <T> AggregatedPage<SearchHit<T>> moreLikeThis(MoreLikeThisQuery query, Class<T> clazz,
```
mp911de: How does `AggregatedPage` relate to `SearchHit`/`SearchHits`? Can there be an `AggregatedPage` that contains `T` instead of `SearchHit<T>`?

Related (in the context of a new ticket): Would it make sense to rename `AggregatedPage` to `AggregationPage` and let a deprecated `AggregatedPage` extend from `AggregationPage`?

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -135,9 +138,10 @@ public void delete(Query query, Class<?> clazz, IndexCoordinates index) {
 		return doMultiSearch(queries, classes, request);
 	}
 
-	private <T> List<Page<T>> doMultiSearch(List<? extends Query> queries, Class<T> clazz, MultiSearchRequest request) {
+	private <T> List<Page<SearchHit<T>>> doMultiSearch(List<? extends Query> queries, Class<T> clazz,
+			MultiSearchRequest request) {
 		MultiSearchResponse.Item[] items = getMultiSearchResult(request);
-		List<Page<T>> res = new ArrayList<>(queries.size());
+		List<Page<SearchHit<T>>> res = new ArrayList<>(queries.size());
```
mp911de: How about a new ticket that introduces a `MultiResult<T>` type? `List<Page<SearchHit<T>>>` is somewhat hard to read.

src/main/java/org/springframework/data/elasticsearch/core/SearchHit.java
```diff
@@ -0,0 +1,76 @@
...
+ * Encapsulates the found data with additional information from the search.
+ * 
+ * @param <T> the result data class.
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public final class SearchHit<T> {
+
+	private final float score;
+	private final T hit;
```
mp911de: Nit: `T content`

src/main/java/org/springframework/data/elasticsearch/core/SearchHit.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * Encapsulates the found data with additional information from the search.
+ * 
+ * @param <T> the result data class.
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public final class SearchHit<T> {
```
mp911de: Let's also add a `toString` method, see [`GeoResult`](https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/geo/GeoResult.java#L44) for some inspiration.

Does it make sense to also associated the document Id with the `SearchHit` as the identifier has a fixed ES type?

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -81,7 +81,7 @@ default long count(Query query, IndexCoordinates index) {
 	 * @param index the index to run the query against
 	 * @return a page with aggregations
 	 */
-	<T> AggregatedPage<T> queryForPage(Query query, Class<T> clazz, IndexCoordinates index);
+	<T> AggregatedPage<SearchHit<T>> queryForPage(Query query, Class<T> clazz, IndexCoordinates index);
```
mp911de: Related to the question above: Changing the component type to `SearchHit<T>` could make it hard for users to consume the actual result. We might want to add convenience methods to `AggregatedPage` to consume a `List<T>` ( `List<T> unwrap()`?). Just a thought.

src/test/java/org/springframework/data/elasticsearch/core/geo/ElasticsearchTemplateGeoTests.java
```diff
@@ -172,7 +173,7 @@ public void shouldFindStringAnnotatedGeoMarkersInRangeForGivenCriteriaQuery() {
 		CriteriaQuery geoLocationCriteriaQuery = new CriteriaQuery(
 				new Criteria("locationAsString").within(new GeoPoint(51.000000, 0.100000), "1km"));
 		// when
-		List<LocationMarkerEntity> geoAuthorsForGeoCriteria = operations.queryForList(geoLocationCriteriaQuery,
+		List<SearchHit<LocationMarkerEntity>> geoAuthorsForGeoCriteria = operations.queryForList(geoLocationCriteriaQuery,
```
mp911de: Once we have `SearchHits`, the migration will be likely easier when users can switch from `List<T>` to `SearchHits<T>`.

src/test/java/org/springframework/data/elasticsearch/core/query/CriteriaQueryTests.java
```diff
@@ -777,7 +778,7 @@ public void shouldPerformBoostOperation() {
 		CriteriaQuery criteriaQuery = new CriteriaQuery(new Criteria("message").contains("foo").boost(1));
 
 		// when
-		Page<SampleEntity> page = operations.queryForPage(criteriaQuery, SampleEntity.class, index);
+		Page<SearchHit<SampleEntity>> page = operations.queryForPage(criteriaQuery, SampleEntity.class, index);
```
mp911de: Would it also make sense to introduce a `SearchPage` type to avoid the deep generic parametrization?

sothawo:
I'll address the "minor" issues later. About the ideas with `SearchHits`, `SearchPage` etc: I agree, but I wouldn't wnat to do this all in this ticket. This should be the first, but definitely not the last step. So I'd prefer to first add the `SearchHit` - to the reactive part as well - and then in following tickets improve the API and user experience in smaller steps

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
sothawo: `AggregatedPage` extends `ScrolledPage>`and `ScoredPage` (I'm not sure if we need these as separate interfaces, as `AggregatedPageImpl` is the only implementaton of it all, and it adds ES aggregations. I don't like this class as it exposes ES objects in the API, but at the moment it's the only way to return aggregations. That's something I'd like to change later.

sothawo:
as for different methods for returning `List<T>` or `List<SearchHit<T>>`: This goes somehow together with the repositories: There are in both reactive and non-reactive repositories the _find_ methods returning `List<T>` and `Flux<T>`. These we cannot change, this comes from spring-data-commons.

The reactive repositories up to now only use _find_ whereas the non-reactive use _find_ and _search_

The reactive operations use _find_, the non-reactive _query_.

Hard to find a way to have as little breaking change as possible.

src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
```diff
@@ -81,4 +88,55 @@ public static Object unwrapSearchHits(Object result) {
...
+
+		@Override
+		@Nullable
+		public String getScrollId() {
+			return searchHits.getScrollId();
+		}
+	}
+
+	private static Pageable pageableOrUnpaged(Pageable pageable) {
+		return ofNullable(pageable).orElse(Pageable.unpaged());
```
xhaggi: Why not a simple null check?

src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
```diff
@@ -81,4 +88,55 @@ public static Object unwrapSearchHits(Object result) {
 
 		return result;
 	}
+
+	public static <T> SearchPage<T> searchPageFor(SearchHits<T> searchHits, @Nullable Pageable pageable) {
```
xhaggi: `toSearchPage()`, IMO sound better.

src/main/java/org/springframework/data/elasticsearch/core/SearchPage.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import org.springframework.data.elasticsearch.core.aggregation.AggregatedPage;
+
+/**
+ * An aggregated page that holds a {@link SearchHits} element.
+ * 
+ * @author Peter-Josef Meisch
```
xhaggi: `@since 4.0`?

src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
sothawo: I took this from `AggregatePageImpl` because it's a nice one-liner. The null check version would be:
```
	private static Pageable pageableOrUnpaged(Pageable pageable) {
		if(pageable != null) {
			return pageable;
		} else {
			return Pageable.unpaged();
		}
        }
```

src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
sothawo: For me, methods named _toSomething()_ transform the object that they are called upon into something different. Here we have a static utility function so I deliberately did not use _to.._

sothawo:
Got quite further than I thought. We now have:

- `SearchHit<T>` enriching  a T with _id_ and _score_
- `SearchHits<T>` for a search result, containing _totalHits_, _maxScore_,_scrollId_ and a list of `SearchHit<T>`.
- `SearchPage<T>` as a Page implementations with a `SearchHits<T>` and replacing the previous interfaces `AggregatePage`, `ScrolledPage` and `ScoredPage`.

The Operations and Templates now use these classes as search results; the repositories return the _normal_ `List<T>` and `Page<T>` as before.

This should give us a solid base to add more and more information from the search like highlights, sorts etc.

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQuery.java
```diff
@@ -71,17 +72,23 @@ public Object execute(Object[] parameters) {
 		StringQuery stringQuery = createQuery(accessor);
```
mp911de: In a later stage, we should move the query execution into a strategy object (see [`MongoQueryExecution`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/MongoQueryExecution.java#L53) to execute the query according to what we want to see as results (see also [MongoDB](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/AbstractMongoQuery.java#L129-L156)).

Unconditionally calling `unwrapSearchHits` seems a workaround for a missing execution model.

sothawo: Just a short reply before I find more time:

- I think the aggregations are up to now contained in the `AggregatedPage` because that was the only possibility to return additional information for a search result. From the ES concept, aggregations are an addition to a search result, so they should be moved to the `SearchHits` class (and I would not keep the `AggregatedPage`.
- Renaming from `query..()` to `search..()`: yes, but I would not mix this with this PR but do a follow up on this.
- I'll check the `scroll` methods and see how adding `SearchSlice` makes this cleaner, probably we should separate the scrolling from the paging.
- Totally replacing page with scroll would probably make sense as ES doesn't have the concept of pages and scrolling is faster than searching with offset and size, but that would have breaking changes for the repository interface as well, as up to now repositories can return pages and not the scrollid, and have to check if the scrollid is supported in the `Query`implementations.

Up to now the breaking changes are in the `SearchOperations` implementation with the repositories working as before; I'd like to keep that this way for now.

mp911de: What do you think about reducing this PR to introducing the new `SearchHit`/`SearchHits`/… types only? I agree that the method changes are a better fit for a separate PR.

sothawo: that would be commit f4203f9 plus adding some info to the `SearchHits` plus some more info to the class. I'll get that running and make further tickets for changing the methods, moving aggregates from the page to SearchHits etc.

xhaggi: >Totally replacing page with scroll would probably make sense as ES doesn't have the concept of pages and scrolling is faster than searching with offset and size, but that would have breaking changes for the repository interface as well, as up to now repositories can return pages and not the scrollid, and have to check if the scrollid is supported in the Query implementations.

Scrolling is a stateful operation in ES, while searching is stateless. Using Scroll as the default for repo methods that return a page is not a good idea.

mp911de: > Totally replacing page with scroll would probably make sense

I was referring to `startScroll` and `continueScroll` methods which return a `Page` object. Both should rather return a `Slice` type. Repository methods are not affected by this change.

xhaggi: Ok sorry for the noise here.

sothawo:
I removed the changes that went beyond the `SearchHit`and `SearchHits`.

src/main/java/org/springframework/data/elasticsearch/core/SearchHits.java
```diff
@@ -0,0 +1,98 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * Encapsulates a list of {@link SearchHit}s with additional information from the search.
+ *
+ * @param <T> the result data class.
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public class SearchHits<T> implements Iterable<SearchHit<T>> {
```
mp911de: We could implement `Streamable` instead of `Iterable` so we would inherit improved `Stream` support out of the box.

sothawo:
changed from `Iterable`to `Streamable and changed `SearchOperations` so the all the methods that return `SearchHit(s)` now have names starting with _search_ , the old methods are unchanged in their signature but deprecated and they now just use the _search_ methods.

Still need to introduce the new functions to the reactive part - a task for tomorrow.

sothawo:
now the `ReactiveSearchOperations` as well has the _find_ methods with the old return types deprecated and has _search_ methods returning the `SearchHit`variants.

That was a good idea to do it this way instead of renamimg them later, thx Mark

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
-	<T> ScrolledPage<T> continueScroll(@Nullable String scrollId, long scrollTimeInMillis, Class<T> clazz);
+	/**
+	 * Returns next scrolled page
+	 *
+	 * @param scrollId the scroll id
+	 * @param scrollTimeInMillis duration of the scroll time
+	 * @param clazz The class of entity to retrieve.
+	 * @return scrolled page result
+	 */
+	<T> ScrolledPage<SearchHit<T>> searchScrollContinue(@Nullable String scrollId, long scrollTimeInMillis,
```
mp911de: `startScroll`/`continueScroll` methods are rather internal methods and we expose a streaming functionality on top of these. Once we have `SearchSlice` in place, we should be good to just change the return types of `startScroll`/`continueScroll` so there's no need to have `searchScrollContinue`/`searchScrollClear`.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+
+	/**
+	 * Execute the query against elasticsearch and return the first returned object.
+	 *
+	 * @param query the query to execute
+	 * @param clazz the entity clazz used for property mapping
+	 * @param index the index to run the query against
+	 * @return the first matching object
+	 */
+	default <T> SearchHit<T> searchForObject(Query query, Class<T> clazz, IndexCoordinates index) {
```
mp911de: Let's stick with `searchOne` to align with the `findOne`/`selectOne` naming scheme that we have in the other modules.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+	/**
+	 * Execute the criteria query against elasticsearch and return result as {@link List}
+	 *
+	 * @param <T> element return type
+	 * @param query the query to execute
+	 * @param clazz the entity clazz used for property mapping
+	 * @param index the index to run the query against
+	 * @return list of found objects
+	 */
+	<T> SearchHits<T> searchForList(Query query, Class<T> clazz, IndexCoordinates index);
```
mp911de: Let's stick with just `search` to align with Spring Data naming.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+
+	/**
+	 * Execute the multi-search against elasticsearch and return result as {@link List} of {@link Page}
+	 *
+	 * @param queries the queries
+	 * @param clazz the entity clazz used for property mapping
+	 * @param index the index to run the query against
+	 * @return list of pages with the results
+	 */
+	<T> List<Page<SearchHit<T>>> searchForPage(List<? extends Query> queries, Class<T> clazz, IndexCoordinates index);
```
mp911de: Let's rename this method to `multiSearch`. I also wonder what's the benefit of having `List<Page<…>>` over `List<SearchHits<T>>` in the return type.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+
+	/**
+	 * Execute the multi-search against elasticsearch and return result as {@link List} of {@link Page}
+	 *
+	 * @param queries the queries
+	 * @param classes the entity classes used for the queries
+	 * @param index the index to run the query against
+	 * @return list of pages with the results
+	 */
+	List<AggregatedPage<? extends SearchHit<?>>> searchForPage(List<? extends Query> queries, List<Class<?>> classes,
```
mp911de: Let's rename this method to `multiSearch`. I also wonder what's the benefit of having `List<Page<…>>` over `List<SearchHits<T>>` in the return type.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+	/**
+	 * Execute the multi search query against elasticsearch and return result as {@link List}
+	 *
+	 * @param queries the queries to execute
+	 * @param clazz the entity clazz used for property mapping
+	 * @param index the index to run the query against
+	 * @param <T> element return type
+	 * @return list of found objects
+	 */
+	default <T> List<List<SearchHit<T>>> searchForList(List<Query> queries, Class<T> clazz, IndexCoordinates index) {
```
mp911de: If we rename `searchForPage(List, …)` to `multiSearch` and align the result to `List<SearchHits<T>>`, there's no need for this method, right?

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+
+	/**
+	 * Execute the multi search query against elasticsearch and return result as {@link List}
+	 *
+	 * @param queries the queries to execute
+	 * @param classes the entity classes used for property mapping
+	 * @param index the index to run the query against
+	 * @return list of list of found objects
+	 */
+	default List<List<?>> searchForList(List<Query> queries, List<Class<?>> classes, IndexCoordinates index) {
```
mp911de: If we rename `searchForPage(List, …)` to `multiSearch` and align the result to `List<SearchHits<T>>`, there's no need for this method, right?

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+
+	/**
+	 * Execute the query against elasticsearch and return ids
+	 *
 	 * @param query the query to execute
 	 * @param clazz the entity clazz used for property mapping
 	 * @param index the index to run the query against
+	 * @return list of found object ids
+	 */
+	List<String> searchForIds(Query query, Class<?> clazz, IndexCoordinates index);
```
mp911de: Let's not introduce this method as this method creates API surface for something that is easily consumable with `SearchHit.getId()`. Internally, there's no mapping happening and when using `SearchDocument`, we also skip mapping.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+	/**
+	 * Returns scrolled page for given query
+	 *
+	 * @param scrollTimeInMillis duration of the scroll time
+	 * @param query The search query.
+	 * @param clazz The class of entity to retrieve.
+	 * @param index the index to run the query against
+	 * @return scrolled page result
+	 */
+	<T> ScrolledPage<SearchHit<T>> searchScrollStart(long scrollTimeInMillis, Query query, Class<T> clazz,
```
mp911de: How about not introducing these methods yet as we should have `SearchSlice` in place first. See also above my comment that scroll methods have a rather internal flavor and we should be able to just change `startScroll`/`continueScroll` return types.

src/main/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchConverter.java
```diff
@@ -72,14 +75,24 @@ default ProjectionFactory getProjectionFactory() {
...
+
+	/**
+	 * builds a {@link SearchHit} from a {@link SearchDocument}.
+	 * 
+	 * @param searchDocument must not be {@literal null}
+	 * @param type the type of the returned data, must not be {@literal null}.
+	 * @param <T> the clazz of the type, must not be {@literal null}.
+	 * @return SearchHIt with all available information filled in
+	 */
+	<T> SearchHit<T> toSearchHit(SearchDocument searchDocument, Class<T> type);
```
mp911de: Let's align with `EntityReader`'s `read` method so we would have `<R> SearchHit<R> read(Class<R> type, SearchDocument searchDocument);`

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +202,203 @@ default long count(Query query, IndexCoordinates index) {
...
+	 * @param queries the queries
+	 * @param classes the entity classes used for the queries
+	 * @param index the index to run the query against
+	 * @return list of pages with the results
+	 */
+	List<AggregatedPage<? extends SearchHit<?>>> searchForPage(List<? extends Query> queries, List<Class<?>> classes,
+			IndexCoordinates index);
+
+	/**
+	 * Execute the criteria query against elasticsearch and return result as {@link List}
```
mp911de: We should also cleanup the Javadoc before cleanup (`SearchHits` instead of list). Check also the other methods.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
sothawo: The only need of having the page returned instead of the list is that this currently is the only possibility to get aggregations returned. Aggregations don't belong in the page but in the `SearchHits`. I have created a ticket for moving them there, so when we have that implemented, we can consolidate the different methods returning `List` and `Page` and remove the page variants because the page returned does not hold relevant info then. This affects not only _multiSearch_  but the single query _search_ as well.

sothawo: I renamed it to `multiSearchForPage` and changed the overload that retruns a `List<SearchHits<T>>` to `multiSearch`. We should aim to keep the methods returning the `SearchHits` and not the ones with the `AggregatedPage`s once we have the necessary information in the `SearchHits`class.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
sothawo: I changed this to `List<SearchHits<T>> multiSearch(List<Query> queries, Class<T> clazz, IndexCoordinates index)`.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
sothawo: We need the `SearchHit` variant of these for the `public <T> CloseableIterator<SearchHit<T>> searchForStream(Query query, Class<T> clazz, IndexCoordinates index)`method to be able to return a stream of `SearchHit` objects. If we change the return of `startScroll/continueScroll` later, we cannot change the stream now.

sothawo:
Did more refactorings and cleanups according to the review comments.

The scroll stuff I'd like to do together with a `SearchSlice`.

As for the `AggregatedPage` we won't need this anymore when aggregations are moved to the SearchHits; then the `ScoredPage` can be dropped as well as this info already in in the `SearchHits`.

Next steps would be:
- add sortValues from the result to `SearchHit`(implemented already, just waiting for this PR to be done)
- move aggregations to `SearchHits` and get rid of `AggregatedPage` and `ScoredPage`
- introduce `SearchSlice`and refactor scrolling / streaming

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
mp911de: Thanks for the detail. I also felt that aggregations should go into `SearchHits` (or a subtype of it).

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
mp911de: Okay. Let's revisit the `search…Scroll` methods at a later stage, once we have all result and slice types in place.

src/main/java/org/springframework/data/elasticsearch/core/SearchOperations.java
```diff
@@ -170,32 +203,204 @@ default long count(Query query, IndexCoordinates index) {
...
+
+	/**
+	 * Execute the query against elasticsearch and return the first returned object.
+	 *
+	 * @param query the query to execute
+	 * @param clazz the entity clazz used for property mapping
+	 * @param index the index to run the query against
+	 * @return the first found object
+	 */
+	default <T> SearchHit<T> searchOne(Query query, Class<T> clazz, IndexCoordinates index) {
```
mp911de: Nit: `@Nullable` return.

PR #378, opened 2020-01-09
DATAES-661 - Support track_total_hits request parameter.

sothawo:


keeganwitt:
Not sure where best to mention this, so I mentioned here and in [Jira|https://jira.spring.io/browse/DATAES-661], but shouldn't this also be added to `NativeSearchQueryBuilder` with like `withTrackTotalHits()`, so you don't have to manually call the setter after invoking `build()`?

PR #385, opened 2020-02-02
DATAES-587 - Add lifecycle events.

sothawo:
Adding basic callback functionality. Takes the same approach as in spring-data-mongodb and spring-data-cassandra.

The reactive part is still missing, I need feedback first if this is basically ok.

After adding this we could easily add the auditing feature like it is done in the other spring data modules.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/event/BeforeDeleteCallback.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/event/BeforeDeleteCallback.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


src/main/java/org/springframework/data/elasticsearch/core/event/BeforeDeleteCallback.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.data.elasticsearch.core.mapping.IndexCoordinates;
+
+/**
+ * Callback that is called before an entity is deleted from the index.
+ *
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+@FunctionalInterface
+public interface BeforeDeleteCallback extends ElasticsearchEntityCallback<String> {
```
mp911de: We typically provide callbacks only for save and delete usecases. In both cases, the callbacks are tied to entities only.

PR #388, opened 2020-02-09
DATAES-741 / Tests fail due to Elasticsearch cluster 'blocks' on near…

rpuch:
…ly-full file-systems.

If a filesystem available space is lower than 5-10%, even though it may be dozens of gigabytes, Elasticsearch  may block all the indices making them work in read-only mode. In such a case, tests fail with messages like the following:

org.elasticsearch.cluster.block.ClusterBlockException: blocked by: [FORBIDDEN/12/index read-only / allow delete (api)]

The problem is well-known: https://stackoverflow.com/questions/50609417/elasticsearch-error-cluster-block-exception-forbidden-12-index-read-only-all

The idea of the fix is to configure Elasticsearch with absolute (instead of relative, using percentage) watermarks of a sensible size (like 1Gb) that are ok for tests.

- [ ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAES).

### DIFF
--- a/src/test/resources/node-client-configuration.yml
+++ b/src/test/resources/node-client-configuration.yml
@@ -1,3 +1,6 @@
 #enabled scripts - this require groovy
 #script.inline: true
 #node.max_local_storage_nodes: 100
+cluster.routing.allocation.disk.watermark.low: 1gb
+cluster.routing.allocation.disk.watermark.high: 1gb
+cluster.routing.allocation.disk.watermark.flood_stage: 1gb


sothawo:
please see[see my comment in Jira](https://jira.spring.io/browse/DATAES-741?focusedCommentId=188088&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-188088)

rpuch:
Closing this PR as a new version against master has been created: https://github.com/spring-projects/spring-data-elasticsearch/pull/393

PR #396, opened 2020-02-23
Dataes 745 consolidate operations api

sothawo:
Changes consolidating the Operations API. 

`IndexOperations` now are bound to either an entity or an `IndexCoordinate` object and so know on which index they operate. To get an implementation one needs to call either `ElasticsearchOperations.getIndexOperations(Class)` or ``ElasticsearchOperations.getIndexOperations(IndexCoordinates)`.  There also are methods `IndexOperations.indexOps(Class)` and `IndexOperations.indexOps(IndexCoordinates)` to get a new one for a different index.

Otherwise I added entity based save/delete operations, switched to a consistent naming as discussed.


src/main/java/org/springframework/data/elasticsearch/core/AbstractDefaultIndexOperations.java
```diff
@@ -48,88 +53,158 @@
 	protected final ElasticsearchConverter elasticsearchConverter;
 	protected final RequestFactory requestFactory;
 
-	public AbstractDefaultIndexOperations(ElasticsearchConverter elasticsearchConverter) {
+	@Nullable protected final Class<?> boundClass;
+	@Nullable protected final IndexCoordinates boundIndex;
```
mp911de: `IndexCoordinates` should never be null. Rather, we should derive `IndexCoordinates` from the given class.

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -113,6 +123,24 @@
...
 	<T> List<T> multiGet(Query query, Class<T> clazz, IndexCoordinates index);
 
+	/**
+	 * Check if an entity with given {@literal id} exists.
+	 *
+	 * @param id the {@literal _id} of the document to look for.
+	 * @param clazz the domain type used.
+	 * @return {@literal true} if a matching document exists, {@literal false} otherwise.
+	 */
+	<T> Boolean exists(String id, Class<T> clazz);
```
mp911de: Nit: Use primitive `boolean` to avoid potential null issues. Also, we do not need `T` here. `Class<?>` is fine.

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -113,6 +123,24 @@
...
+
+	/**
+	 * Check if an entity with given {@literal id} exists.
+	 *
+	 * @param id the {@literal _id} of the document to look for.
+	 * @param index the target index, must not be {@literal null}
+	 * @return {@literal true} if a matching document exists, {@literal false} otherwise.
+	 */
+	Boolean exists(String id, IndexCoordinates index);
```
mp911de: Nit: Primitive boolean.

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -158,6 +186,34 @@ default void bulkUpdate(List<UpdateQuery> queries, IndexCoordinates index) {
...
+	String delete(String id, Class<?> entityType);
+
+	/**
+	 * Deletes the given entity
+	 *
+	 * @param entity the entity to delete
+	 * @param <T> the entity class
+	 * @return documentId of the document deleted
+	 */
+	<T> String delete(T entity);
```
mp911de: Nit: `<T>` is not required as we do not make any type contracts here. `delete(Object)` is fine.

src/main/java/org/springframework/data/elasticsearch/core/DocumentOperations.java
```diff
@@ -158,6 +186,34 @@ default void bulkUpdate(List<UpdateQuery> queries, IndexCoordinates index) {
...
+
+	/**
+	 * Deletes the given entity
+	 *
+	 * @param entity the entity to delete
+	 * @param index the index from which to delete
+	 * @param <T> the entity class
+	 * @return documentId of the document deleted
+	 */
+	<T> String delete(T entity, IndexCoordinates index);
```
mp911de: See above, replace `T` with `Object`

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -127,12 +127,11 @@ public String index(IndexQuery query, IndexCoordinates index) {
 	}
 
 	@Override
-	@Nullable
-	public <T> T get(GetQuery query, Class<T> clazz, IndexCoordinates index) {
-		GetRequest request = requestFactory.getRequest(query, index);
+	public <T> Optional<T> get(String id, Class<T> clazz, IndexCoordinates index) {
```
mp911de: Let's stick with `T` instead of `Optional<T>`. A huge number of folks already asked us to roll back the `Optional` change at `findById` and `Optional` isn't being loved that much.

src/main/java/org/springframework/data/elasticsearch/core/IndexOperations.java
```diff
@@ -19,44 +19,181 @@
...
  */
 public interface IndexOperations {
 
+	/**
+	 * returns an {@link IndexOperations} bound to the given entity class.
+	 * 
+	 * @param clazz the entity class
+	 * @return IndexOperations;
+	 */
+	IndexOperations indexOps(Class<?> clazz);
```
mp911de: Let's not introduce `indexOps(…)` methods here. They should live only on `ElasticsearchOperations` only. The deprecated methods that do the forwarding should remain on `ElasticsearchOperations`, too.

src/main/java/org/springframework/data/elasticsearch/core/AbstractDefaultIndexOperations.java
```diff
@@ -48,88 +53,158 @@
...
+		this.elasticsearchConverter = elasticsearchConverter;
+		requestFactory = new RequestFactory(elasticsearchConverter);
+
+		this.boundClass = null;
+		this.boundIndex = boundIndex;
 	}
 
+	protected Class<?> checkForBoundClass() {
+		if (boundClass == null) {
+			throw new IndexOperationsException("IndexOperations are not bound");
```
mp911de: That's rather `InvalidDataAccessApiUsageException`.

src/main/java/org/springframework/data/elasticsearch/core/IndexOperationsException.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+/**
+ * @author Peter-Josef Meisch
+ */
+public class IndexOperationsException extends RuntimeException {
```
mp911de: Let's not introduce this exception. The places that use this exception should rather throw `InvalidDataAccessApiUsageException`.

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
```diff
@@ -151,8 +151,36 @@
...
+
+	/**
+	 * Find the document with the given {@literal id} mapped onto the given {@literal entityType}.
+	 *
+	 * @param id the {@literal _id} of the document to fetch.
+	 * @param entityType the domain type used for mapping the document.
+	 * @param <T>
+	 * @return {@link Mono#empty()} if not found.
+	 */
+	<T> Mono<T> get(String id, Class<T> entityType);
```
mp911de: Nit: `@since 4.0`

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
```diff
@@ -188,7 +227,7 @@
 	 * @param entity must not be {@literal null}.
 	 * @return a {@link Mono} emitting the {@literal id} of the removed document.
 	 */
-	Mono<String> delete(Object entity);
+	<T> Mono<String> delete(T entity);
```
mp911de: `delete(Object)` is fine. We don't need generic types here.

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
```diff
@@ -197,7 +236,7 @@
 	 * @param index the target index, must not be {@literal null}
 	 * @return a {@link Mono} emitting the {@literal id} of the removed document.
 	 */
-	Mono<String> delete(Object entity, IndexCoordinates index);
+	<T> Mono<String> delete(T entity, IndexCoordinates index);
```
mp911de: No need for `T`. Let's stick with `Object`

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
```diff
@@ -220,7 +254,7 @@
 	 * @param entityType must not be {@literal null}.
 	 * @return a {@link Mono} emitting the {@literal id} of the removed document.
 	 */
-	Mono<String> deleteById(String id, Class<?> entityType);
+	Mono<String> delete(String id, Class<?> entityType);
```
mp911de: Nit: `@since 4.0`

src/main/java/org/springframework/data/elasticsearch/core/query/Query.java
```diff
@@ -229,4 +229,29 @@ default Integer getMaxResults() {
...
+	 * @param scrollTimeInMillis the scrolltime settings
+	 * @since 4.0
+	 */
+	void setScrollTimeInMillis(Long scrollTimeInMillis);
+
+	/**
+	 * @return {@literal true} if scrollTimeMillis is set.
+	 * @since 4.0
+	 */
+	default Boolean hasScrollTimeMillis() {
```
mp911de: return `boolean`

src/main/java/org/springframework/data/elasticsearch/core/query/Query.java
```diff
@@ -229,4 +229,29 @@ default Integer getMaxResults() {
...
+	@Nullable
+	Long getScrollTimeInMillis();
+
+	/**
+	 * For queries that are used in delete request, these are internally handled by Elasticsearch as scroll/bulk delete queries.
+	 * 
+	 * @param scrollTimeInMillis the scrolltime settings
+	 * @since 4.0
+	 */
+	void setScrollTimeInMillis(Long scrollTimeInMillis);
```
mp911de: How about changing the base property to `ScrollTime(Duration)`? `inMillis` is a quite dated pattern before we had proper duration types.

src/main/java/org/springframework/data/elasticsearch/core/query/UpdateResponse.java
```diff
@@ -0,0 +1,36 @@
...
+package org.springframework.data.elasticsearch.core.query;
+
+/**
+ * @author Peter-Josef Meisch
+ */
+public class UpdateResponse {
+
+	private Result result;
+
+	public UpdateResponse(Result result) {
```
mp911de: Nit: Non-null assertion. This class should get also a bit of documentation.

src/main/java/org/springframework/data/elasticsearch/repository/support/AbstractElasticsearchRepository.java
```diff
@@ -203,15 +195,19 @@ public long count() {
 
 		Assert.notNull(entities, "Cannot insert 'null' as a List.");
 
-		operations.save(entities, getIndexCoordinates());
-		indexOperations.refresh(getIndexCoordinates());
+		Iterator<S> iterator = entities.iterator();
+		if (iterator.hasNext()) {
+			S entity = iterator.next();
+			operations.save(entities, getIndexCoordinates());
```
mp911de: How about returning `List<T>` from `save(…)` to investigate the index coordinates? 
In some cases, we've seen folks using `repo.saveAll(() -> Stream.of(…).map(…))` which isn't a proper iterable but it's used that way.

sothawo:
fixed all the issues. `IndexOperations`now are always bound, they can only be created by an `ElasticsearchOperations` instance.

PR #401, opened 2020-03-03
DATAES-751 - Introduce ClientCallback for the rest client.

sothawo:


src/main/java/org/springframework/data/elasticsearch/NoSuchIndexException.java
```diff
@@ -15,13 +15,11 @@
...
 package org.springframework.data.elasticsearch;
 
-import org.springframework.dao.NonTransientDataAccessResourceException;
-
 /**
  * @author Christoph Strobl
  * @since 3.2
  */
-public class NoSuchIndexException extends NonTransientDataAccessResourceException {
+public class NoSuchIndexException extends SpringDataElasticsearchException {
```
mp911de: That exception was already fine. No need to turn this one into an uncategorized exception.

src/main/java/org/springframework/data/elasticsearch/SpringDataElasticsearchException.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.springframework.data.elasticsearch;
+
+import org.springframework.dao.UncategorizedDataAccessException;
+
+/**
+ * @author Peter-Josef Meisch
+ * @since 4.0
+ */
+public class SpringDataElasticsearchException extends UncategorizedDataAccessException {
```
mp911de: Instead of adding another exception type, let's throw `UncategorizedDataAccessException` in the places that we're not fully sure about the cause.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -406,4 +407,26 @@ default String stringIdRepresentation(@Nullable Object id) {
...
+
+	/**
+	 * Callback interface to be used with {@link #execute(RestHighLevelClientCallback)} for operating directly on
+	 * {@link org.springframework.data.elasticsearch.client.reactive.ReactiveElasticsearchClient}.
+	 * 
+	 * @since 4.0
+	 */
+	@FunctionalInterface
+	interface RestHighLevelClientCallback<T> {
+		T doWithClient(RestHighLevelClient client) throws Throwable;
```
mp911de: Declaring a checked exception such as `IOException` is fine but we must not generalize towards `Throwable`.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -406,4 +407,26 @@ default String stringIdRepresentation(@Nullable Object id) {
...
+	/**
+	 * Execute a callback with the {@link RestHighLevelClient}
+	 * 
+	 * @param callback the callback to execute, must not be {@literal null}
+	 * @param <T> the type returned from the callback
+	 * @return the callback result
+	 * @throws UnsupportedOperationException for the TransportClient based implementation.
+	 * @since 4.0
+	 */
+	<T> T execute(RestHighLevelClientCallback<T> callback) throws RuntimeException;
```
mp911de: Let's expose this method on the REST Template only for a more concise applicability. Teasing client callback and then explaining that it does not work for the Transport Client is confusing.

The thrown type should be `DataAccessException`.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -339,26 +282,27 @@ public SearchResponse suggest(SuggestBuilder suggestion, IndexCoordinates index)
...
 	// endregion
+
+	@Override
+	public <T> T execute(RestHighLevelClientCallback<T> callback) throws RuntimeException {
+
+		Assert.notNull(callback, "callback must not be null");
+
+		try {
+			return callback.doWithClient(client);
+		} catch (Throwable t) {
```
mp911de: Limit the caught exceptions to the actual ones such as `RuntimeException` and `IOException`.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchExceptionTranslator.java
```diff
@@ -16,25 +16,48 @@
...
 public class ElasticsearchExceptionTranslator implements PersistenceExceptionTranslator {
 
+	/**
+	 * translates a Throwable. Exceptions that are no {@link RuntimeException}s are wrapped in a RuntimeException
+	 * 
+	 * @param throwable the Throwable to map
+	 * @return the translated RuntimeException.
+	 * @since 4.0
+	 */
+	public RuntimeException translateException(Throwable throwable) {
```
mp911de: Let's stick with the `translateExceptionIfPossible` pattern that returns a potentially nullable `DataAccessException`. Bubbling of the original exception happens then on Template API level.

src/main/java/org/springframework/data/elasticsearch/SpringDataElasticsearchException.java
sothawo: `UncategorizedDataAccessException` is abstract, that's why I introduced this class

mp911de: That intent wasn’t obvious from the name and usage. Then let’s rename it to UncategorizedElasticaearchException for a clearer picture.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchExceptionTranslator.java
sothawo: this is code I just moved from [`ReactiveElasticsearchTemplate`](https://github.com/spring-projects/spring-data-elasticsearch/blob/3.2.x/src/main/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchTemplate.java#L776-L783).
If `translateException` would return null for an exception that is not mapped to a `DataAccessException`, then the reactive code would create a `MonoError(null)` which  will result in an NPE.
So for me the logic to try to translate the exception - and, if that cannot be done, return it wrapped in a `RuntimeException`if it is not already one, makes sense.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
sothawo: that would mean that the `ClientCallback.doWithClient()` method could only throw IOException as a checked one. Can we put this limitation on that callback method?

sothawo:
some questions regarding the translation mapping, did the other changes

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
mp911de: Yes, that change would indeed make a lot of sense. Implementors aren't required to throw exceptions so we should be good to extend the list of potentially thrown exceptions if we recognize the need for it.

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchExceptionTranslator.java
mp911de: Sorry, I missed to explain that callers of `translateExceptionIfPossible` are required to check for the exception translation outcome to prevent NPEs. Let's keep that method in the Template implementations for now.

sothawo:
cleaned up `ElasticsearchExceptionTranslator` and the handled exceptions

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java
```diff
@@ -339,26 +283,66 @@ public SearchResponse suggest(SuggestBuilder suggestion, IndexCoordinates index)
...
+
+	/**
+	 * translates an Exception if possible. Exceptions that are no {@link RuntimeException}s are wrapped in a
+	 * RuntimeException
+	 *
+	 * @param exception the Exception to map
+	 * @return the potentially translated RuntimeException.
+	 * @since 4.0
+	 */
+	public RuntimeException translateException(Exception exception) {
```
mp911de: Nit: Let's have this method private unless someone asks for it. It's an utility to the client callback.

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
```diff
@@ -261,8 +227,40 @@ protected void doRefresh(IndexCoordinates index) {
...
 			return aliasMetaDataList;
 		} catch (IOException e) {
-			throw new ElasticsearchException("Could not map alias response : " + aliasResponse, e);
+			throw new UncategorizedElasticsearchException("Could not map alias response : " + aliasResponse, e);
+		}
+	}
+	// endregion
+
+	// region clientcallback
+	public <T> T execute(ElasticsearchRestTemplate.ClientCallback<T> callback) throws RuntimeException {
```
mp911de: Can we make the `DefaultIndexOperations` to use the REST Template API? That way, we can reuse the client callback and exception translation without us duplicating the code.

src/main/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchTemplate.java
```diff
@@ -812,4 +803,22 @@ protected ReactiveElasticsearchClient getClient() {
...
 
+	/**
+	 * translates an Exception if possible. Exceptions that are no {@link RuntimeException}s are wrapped in a
+	 * RuntimeException
+	 *
+	 * @param throwable the Throwable to map
+	 * @return the potentially translated RuntimeException.
+	 * @since 4.0
+	 */
+	public RuntimeException translateException(Throwable throwable) {
```
mp911de: Nit: Let's have this method private unless someone asks for it.

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
sothawo: great idea; by doing this I can remove the RestClient from the `DefaultIndexOperations`

PR #405, opened 2020-03-16
Update MappingElasticsearchConverter.java

gsrinivas10:
If map entry value is null, then we are getting null pointer exception  in method 
private boolean isSimpleType(Object value) {
		return isSimpleType(value.getClass());
}

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
@@ -354,6 +354,9 @@ public class MappingElasticsearchConverter
 		Collection<Object> target = createCollectionForValue(targetType, source.size());
 
 		for (Object value : source) {
+			
+			if(null == value)
+				return null;
 
 			if (isSimpleType(value)) {
 				target.add(
--- a/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
@@ -390,12 +390,9 @@ public class MappingElasticsearchConverter
 		Map<String, Object> target = new LinkedHashMap<>();
 		for (Entry<String, Object> entry : source.entrySet()) {
 			
-			if(null == entry.getValue()) {
+			if(entry.getValue() == null) {
 				target.put(entry.getKey(),null);
-				continue ;
-			}
-
-			if (isSimpleType(entry.getValue())) {
+			} else if (isSimpleType(entry.getValue())) {
 				target.put(entry.getKey(),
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@gsrinivas10 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=405)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=405) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gsrinivas10 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=405)!

sothawo:
thank you for this PR.

please create an issue in [Jira](https://jira.spring.io/projects/DATAES/issues) if you think you have found a bug.

As to your PR: the branch you want to merge has failing tests (`MappingElasticsearchConverterUnitTests.writeSubTypeCorrectly`) because the converter does not write null valued fields and does not expect them.

So please use the issue in Jira to describe the problem you have and why you think that handling null values should be handled differently than they are as of now

gsrinivas10:
Hi @sothawo thanks for spending your time and looking into this pull request. I've created a jira ticket https://jira.spring.io/browse/DATAES-763 and attached a sample test file to replicate the issue. Please review it.

src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
```diff
@@ -354,6 +354,9 @@ public void afterPropertiesSet() {
 		Collection<Object> target = createCollectionForValue(targetType, source.size());
 
 		for (Object value : source) {
+			
+			if(null == value)
+				return null;
```
sothawo:
```suggestion
			if(value == null) {
				return null;
			}
		
```
We do the comparison the other way round and always use the curly braces.

src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
```diff
@@ -386,6 +389,11 @@ public void afterPropertiesSet() {
 
 		Map<String, Object> target = new LinkedHashMap<>();
 		for (Entry<String, Object> entry : source.entrySet()) {
+			
+			if(null == entry.getValue()) {
+				target.put(entry.getKey(),null);
+				continue ;
+			}
 
 			if (isSimpleType(entry.getValue())) {
```
sothawo:
```suggestion
			
			if(entry.getValue() == null) {
				target.put(entry.getKey(),null);
			} else if (isSimpleType(entry.getValue())) {
```
rather than using continue, go on with an else if

gsrinivas10:
@sothawo I've added two new test cases and changed the logic of MappingElasticsearchConverter as per your suggesion.

PR #407, opened 2020-03-20
DATAES-766 - Replace CloseableIterator with SearchHitsIterator in stream operations.

xhaggi:
@mp911de @sothawo mind taking a look.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/ScrolledPage.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ScrolledPage.java
@@ -2,13 +2,14 @@
 package org.springframework.data.elasticsearch.core;
 
 import org.springframework.data.domain.Page;
-import org.springframework.lang.Nullable;
 
 /**
  * @author Artur Konczak
  * @author Peter-Josef Meisch
  * @author Sascha Woo
+ * @deprecated since 4.0, will be removed in a future version.
  */
+@Deprecated
 public interface ScrolledPage<T> extends Page<T> {
 
 	String getScrollId();
--- a/src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
@@ -32,6 +32,7 @@ import org.springframework.lang.Nullable;
  * Utility class with helper methods for working with {@link SearchHit}.
  * 
  * @author Peter-Josef Meisch
+ * @author Sascha Woo
  * @since 4.0
  */
 public final class SearchHitSupport {
@@ -95,10 +96,17 @@ public final class SearchHitSupport {
 	 * @param searchHits, must not be {@literal null}.
 	 * @param pageable, must not be {@literal null}.
 	 * @return the created Page
+	 * @deprecated since 4.0, will be removed in a future version.
 	 */
+	@Deprecated
 	public static <T> AggregatedPage<SearchHit<T>> page(SearchHits<T> searchHits, Pageable pageable) {
-		return new AggregatedPageImpl<>(searchHits.getSearchHits(), pageable, searchHits.getTotalHits(),
-				searchHits.getAggregations(), searchHits.getScrollId(), searchHits.getMaxScore());
+		return new AggregatedPageImpl<>( //
+				searchHits.getSearchHits(), //
+				pageable, //
+				searchHits.getTotalHits(), //
+				searchHits.getAggregations(), //
+				null, //
+				searchHits.getMaxScore());
 	}
 
 	public static <T> SearchPage<T> searchPageFor(SearchHits<T> searchHits, @Nullable Pageable pageable) {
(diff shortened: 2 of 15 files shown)


src/main/java/org/springframework/data/elasticsearch/core/SearchHitsIterator.java
```diff
@@ -0,0 +1,34 @@
...
+ * @param <T>
+ * @since 4.0
+ */
+public interface SearchHitsIterator<T> extends CloseableIterator<SearchHit<T>> {
+
+	/**
+	 * @return the number of total hits.
+	 */
+	long getTotalHits();
+}
```
sothawo: what about other information like the total hits relation or the max score (check what's in the `SearchHits` class)?

xhaggi: The scroll methods just return `ScrolledPage` at the moment.

sothawo: wouldn't it make sense to rework the `ScrolledPage` stuff then as well?  
Looking at the rest client (https://github.com/spring-projects/spring-data-elasticsearch/blob/master/src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplate.java#L260-L269) the `SearchDocumentResponse` passed into the converter has all the information.

In the mapper (https://github.com/spring-projects/spring-data-elasticsearch/blob/master/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java#L141-L168) you see just below the method used to return the deprecated `AggregatedPageImpl` how the `SearchHits`object is built.

So maybe this would be a good chance to clean up the methods here a little bit

xhaggi: Sure, will take a look at it later.

sothawo: btw `SearchHits` already has a property for the scrollid, and `SearchPage` wraps a `SearchHits`. So these eventually can be reused for changing the internal implementation of `StreamQueries`

xhaggi:
@sothawo I'm done with the rewrite. The `ScrolledPage` is now deprecated and could be removed in a future version. `SearchHits<T>` is now an interface and no longer expose a `scrollId`. Therefore I added a second interface `SearchScrollHits<T>` for internal use only, which extends `SearchHits<T>`. The basic implementation are in `SearchHitsImpl`. In this way, the scroll context and scroll ID is completely hidden.

It is best to look at the changes ;)

src/main/java/org/springframework/data/elasticsearch/core/ScrolledPage.java
```diff
@@ -2,13 +2,14 @@
 package org.springframework.data.elasticsearch.core;
 
 import org.springframework.data.domain.Page;
-import org.springframework.lang.Nullable;
 
 /**
  * @author Artur Konczak
  * @author Peter-Josef Meisch
  * @author Sascha Woo
+ * @deprecated will be removed in a future version.
```
sothawo:
```suggestion
 * @deprecated since 4.0, will be removed in a future version.
```

src/main/java/org/springframework/data/elasticsearch/core/SearchHitSupport.java
```diff
@@ -97,8 +98,13 @@ public static Object unwrapSearchHits(Object result) {
 	 * @return the created Page
 	 */
 	public static <T> AggregatedPage<SearchHit<T>> page(SearchHits<T> searchHits, Pageable pageable) {
```
sothawo: Let's deprecate this method as well

src/main/java/org/springframework/data/elasticsearch/core/StreamQueries.java
```diff
@@ -35,25 +36,30 @@
 	/**
 	 * Stream query results using {@link ScrolledPage}.
```
sothawo:
```suggestion
	 * Stream query results using {@link SearchScrollHits}.
```

src/test/java/org/springframework/data/elasticsearch/core/StreamQueriesTest.java
```diff
@@ -36,42 +38,51 @@
...
+				scrollId -> {
+				});
+
+		// then
+		assertThat(iterator.getTotalHits()).isEqualTo(1);
+
+	}
+
+	private SearchScrollHits<String> newSearchScrollHits(List<SearchHit<String>> hits) {
+		return new SearchHitsImpl<String>(1, TotalHitsRelation.EQUAL_TO, 0, "1234", hits, null);
```
sothawo:
```suggestion
		return new SearchHitsImpl<String>(hits.size(), TotalHitsRelation.EQUAL_TO, 0, "1234", hits, null);
```

PR #411, opened 2020-03-25
DATAES-653 - Make it easier to use a custom request converter when ex…

rpuch:
…tending DefaultReactiveElasticsearchClient.

- convert RequestConverters class (with static methods) to RequestConverter interface (with default methods)
- use a default RequestConverter implementation in DefaultReactiveElasticsearchClient
- make it possible to supply a custom RequestConverter implementation to be used by DefaultReactiveElasticsearchClient to make it possible to customize how the requests are being built

This is a prototype to try the approach suggested in https://jira.spring.io/browse/DATAES-653

To be honest, I don't like the result too much. The problems are:

* A ton of default methods simply do not seem right. Maybe it's just that I'm not accustomed with interfaces having so much code; probably it's a matter of habit
* In this approach, it's natural to convert static methods to instance methods, but in Java 8 interfaces cannot have private instance methods, so all private methods need to remain static. The best I could do is move them to a separate class (`DefaultRequestConverters`), but here comes another problem: it's not clear what static methods need to remain static and which need to be accessible (package local? public?)

Maybe I just did not understand the idea of @sothawo (mentioned in DATAES-653) correctly.

But it seems that it would look better if we leave the interface (`RequestConverter`) an interface (with only abstract methods) and move all the code to `DefaultRequestConverter`. (In the future, when new methods would need to be added to the interface, they could be added as default methods to avoid breaking compilation of users code).

What do you think?

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
@@ -79,7 +79,6 @@ import org.elasticsearch.action.support.master.AcknowledgedResponse;
 import org.elasticsearch.action.update.UpdateRequest;
 import org.elasticsearch.action.update.UpdateResponse;
 import org.elasticsearch.client.Request;
-import org.elasticsearch.client.core.CountRequest;
 import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.common.xcontent.DeprecationHandler;
 import org.elasticsearch.common.xcontent.NamedXContentRegistry;
@@ -94,14 +93,12 @@ import org.elasticsearch.search.Scroll;
 import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.search.SearchHits;
 import org.reactivestreams.Publisher;
-import org.springframework.data.elasticsearch.ElasticsearchException;
 import org.springframework.data.elasticsearch.client.ClientConfiguration;
 import org.springframework.data.elasticsearch.client.ClientLogger;
 import org.springframework.data.elasticsearch.client.ElasticsearchHost;
 import org.springframework.data.elasticsearch.client.NoReachableHostException;
 import org.springframework.data.elasticsearch.client.reactive.HostProvider.Verification;
 import org.springframework.data.elasticsearch.client.reactive.ReactiveElasticsearchClient.Indices;
-import org.springframework.data.elasticsearch.client.util.RequestConverters;
 import org.springframework.data.util.Lazy;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
@@ -128,6 +125,7 @@ import org.springframework.web.reactive.function.client.WebClient.RequestBodySpe
  * @author Peter-Josef Meisch
  * @author Huw Ayling-Miller
  * @author Henrique Amaral
+ * @author Roman Puchkovskiy
  * @since 3.2
  * @see ClientConfiguration
  * @see ReactiveRestClients
@@ -135,6 +133,8 @@ import org.springframework.web.reactive.function.client.WebClient.RequestBodySpe
 public class DefaultReactiveElasticsearchClient implements ReactiveElasticsearchClient, Indices {
 
 	private final HostProvider hostProvider;
+	
+	private final RequestCreator requestCreator;
 
 	/**
 	 * Create a new {@link DefaultReactiveElasticsearchClient} using the given {@link HostProvider} to obtain server
@@ -143,10 +143,23 @@ public class DefaultReactiveElasticsearchClient implements ReactiveElasticsearch
 	 * @param hostProvider must not be {@literal null}.
 	 */
 	public DefaultReactiveElasticsearchClient(HostProvider hostProvider) {
+		this(hostProvider, new DefaultRequestCreator());
+	}
+
+	/**
+	 * Create a new {@link DefaultReactiveElasticsearchClient} using the given {@link HostProvider} to obtain server
+	 * connections and the given {@link RequestCreator}.
+	 *
+	 * @param hostProvider must not be {@literal null}.
+	 * @param requestCreator must not be {@literal null}.
+	 */
+	public DefaultReactiveElasticsearchClient(HostProvider hostProvider, RequestCreator requestCreator) {
 
 		Assert.notNull(hostProvider, "HostProvider must not be null");
+		Assert.notNull(requestCreator, "RequestCreator must not be null");
 
 		this.hostProvider = hostProvider;
+		this.requestCreator = requestCreator;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


sothawo:
I'll have a look a this probably towards the weekend; won't find the time before that

sothawo:
I wrote:
> We would still need a way to pass this custom request creator to the client

This would rather be a new method in `ReactiveRestClients`.

rpuch:
Hello Peter-Josef, thank you for your review.
Oh, it seems that I misunderstood your original intention.

There are two points that I'd like to clarify.

First, you suggest to make `DefaultRequestCreator` package private, so it would not be useable for the clients. But the clients will usually just need to change the behavior of 1 or 2 methods. Do I understand you correctly that you suggest to implement all `RequestCreator` methods as default methods and keep `DefaultRequestCreator` empty?

Second, on your idea about configuring the `DefaultReactiveElasticsearchClient`. Do I understand you correctly that you suggest to leave `ClientConfiguration` unchanged and add another method in `ReactiveRestClients` with a signature like `public static ReactiveElasticsearchClient create(ClientConfiguration clientConfiguration, RequestCreator creator)`?

Also, is setter ok to inject `RequestCreator` to `DefaultReactiveElasticsearchClient`, or you'd prefer a constructor parameter? I chose a setter originally, because it is not mandatory to supply the `RequestCreator`, a default implementation will be used if it is not supplied.

sothawo:
> First, you suggest to make `DefaultRequestCreator` package private, so it would not be useable for the clients. But the clients will usually just need to change the behavior of 1 or 2 methods. Do I understand you correctly that you suggest to implement all `RequestCreator` methods as default methods and keep `DefaultRequestCreator` empty?

Normally in these cases we make the interface public and the default implementations are package scoped. As in this case the default methods are nearly all just one-liners I think they could be added as default methods to the interface. The `DefaultRequestCreator` would then indeed be just a class implementing the interface without any implementation details of it's own.

> Second, on your idea about configuring the `DefaultReactiveElasticsearchClient`. Do I understand you correctly that you suggest to leave `ClientConfiguration` unchanged and add another method in `ReactiveRestClients` with a signature like `public static ReactiveElasticsearchClient create(ClientConfiguration clientConfiguration, RequestCreator creator)`?

yes, and the existing method `create(ClientConfiguration clientConfiguration)` would just call `create(clientConfiguration, new DefaultRequestCreator())`.

> 
> Also, is setter ok to inject `RequestCreator` to `DefaultReactiveElasticsearchClient`, or you'd prefer a constructor parameter? I chose a setter originally, because it is not mandatory to supply the `RequestCreator`, a default implementation will be used if it is not supplied.

I would go for a constructor parameter that may not be null. `ReactiveElasticsearchClient create(ClientConfiguration clientConfiguration)` would get an additional non null parameter to pass it into the constructor, in  `ReactiveElasticsearchClient create(HttpHeaders headers, String... hosts)` I think we can pass a `new DefaultRequestCreator()` to create.

In this way we can make sure that the request creator is set and not null, when using a setter, the property might be null or not initialized.

rpuch:
Thank you for the clarifications. I've made the corresponding changes. Please let me know if something needs to be changed.

src/main/java/org/springframework/data/elasticsearch/client/reactive/RequestCreator.java
```diff
@@ -0,0 +1,135 @@
...
+import org.springframework.data.elasticsearch.client.util.RequestConverters;
+
+import java.io.IOException;
+import java.util.function.Function;
+
+/**
+ * @author Christoph Strobl
+ * @author Peter-Josef Meisch
+ * @author Roman Puchkovskiy
+ */
```
sothawo:
```suggestion
/**
 * @author Roman Puchkovskiy
 * @ since 4.0
 */
```

src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveRestClients.java
```diff
@@ -41,4 +42,18 @@ public static ReactiveElasticsearchClient create(ClientConfiguration clientConfi
 
 		return DefaultReactiveElasticsearchClient.create(clientConfiguration);
 	}
+
+	/**
+	 * Start here to create a new client tailored to your needs.
+	 *
```
sothawo: add javadoc for the parameters, mentioning that they must not be null (please fix this in the other method as well)

src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
```diff
@@ -143,10 +143,23 @@
 	 * @param hostProvider must not be {@literal null}.
 	 */
 	public DefaultReactiveElasticsearchClient(HostProvider hostProvider) {
+		this(hostProvider, new DefaultRequestCreator());
+	}
+
+	/**
+	 * Create a new {@link DefaultReactiveElasticsearchClient} using the given {@link HostProvider} to obtain server
+	 * connections.
```
sothawo: adapt comment

src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
```diff
@@ -177,8 +190,24 @@ public static ReactiveElasticsearchClient create(HttpHeaders headers, String...
...
+	 * Create a new {@link DefaultReactiveElasticsearchClient} given {@link ClientConfiguration} and
+	 * {@link RequestCreator}. <br />
+	 * <strong>NOTE</strong> If the cluster requires authentication be sure to provide the according {@link HttpHeaders}
+	 * correctly.
+	 *
+	 * @param clientConfiguration Client configuration. Must not be {@literal null}.
+	 * @param requestCreator Request creator. Must not be {@literal null}.
+	 * @return new instance of {@link DefaultReactiveElasticsearchClient}.
+	 */
+	public static ReactiveElasticsearchClient create(ClientConfiguration clientConfiguration,
```
sothawo: ‼️ cannot comment on the line 216 where the `DefaultReactiveElasticsearchClient(hostProvider)`call is made, but there the `requestCreator` constructor argument is missing.

src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultRequestCreator.java
```diff
@@ -0,0 +1,7 @@
+package org.springframework.data.elasticsearch.client.reactive;
+
+/**
+ * @author Roman Puchkovskiy
+ */
```
sothawo:
```suggestion
/**
 * @author Roman Puchkovskiy
 * @since 4.0
 */
```

src/main/java/org/springframework/data/elasticsearch/client/reactive/RequestCreator.java
rpuch: Applied the change manually

src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveRestClients.java
rpuch: Added the javadoc. Please let me know if it's enough.

src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
rpuch: Done

src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
rpuch: Oops, that's embarassing. Fixed.

src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultRequestCreator.java
rpuch: Applied the change manually

rpuch:
Thank you, I've fixed those.

sothawo:
Thanks for the PR, it's merged.

PR #414, opened 2020-03-28
DATAES-771 - Add after-save entity callbacks support.

rpuch:
This change enables to get callbacks when an entity gets saved to Elasticsearch.

Comparing to AfterSaveCallback in spring-data-mongodb (that was used for inspiration), AfterSaveCallback added in this PR has a simpler method signature: it does not have neither `document` (because Elasticsearch does not return an indexed document in a response from index operation) nor `collection` arguments. We can add `index` instead of `collection` though. I need your input on this matter.

Also, I'm going to add support for after-convert callbacks in the next PR if everything goes well with this one.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
@@ -22,6 +22,7 @@ import org.springframework.data.elasticsearch.ElasticsearchException;
 import org.springframework.data.elasticsearch.core.convert.ElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.convert.MappingElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.document.SearchDocumentResponse;
+import org.springframework.data.elasticsearch.core.event.AfterSaveCallback;
 import org.springframework.data.elasticsearch.core.event.BeforeConvertCallback;
 import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentEntity;
 import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentProperty;
@@ -45,6 +46,7 @@ import org.springframework.util.Assert;
  *
  * @author Sascha Woo
  * @author Peter-Josef Meisch
+ * @author Roman Puchkovskiy
  */
 public abstract class AbstractElasticsearchTemplate implements ElasticsearchOperations, ApplicationContextAware {
 
@@ -117,8 +119,13 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 		Assert.notNull(entity, "entity must not be null");
 		Assert.notNull(index, "index must not be null");
 
-		index(getIndexQuery(entity), index);
-		return entity;
+		IndexQuery query = getIndexQuery(entity);
+		index(query, index);
+
+		// suppressing because it's either entity itself or something of a correct type returned by an entity callback
+		@SuppressWarnings("unchecked")
+		T castResult = (T) query.getObject();
+		return castResult;
 	}
 
 	@Override
@@ -151,7 +158,10 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 			});
 		}
 
-		return entities;
+		return indexQueries.stream()
+				.map(IndexQuery::getObject)
+				.map(entity -> (T) entity)
+				.collect(Collectors.toList());
 	}
 
 	@Override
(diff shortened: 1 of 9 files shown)


src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -455,11 +465,39 @@ protected void maybeCallbackBeforeConvertWithQuery(Object query) {
...
-	// don't have a common bas class, therefore the List<?> argument
+	// don't have a common base class, therefore the List<?> argument
 	protected void maybeCallbackBeforeConvertWithQueries(List<?> queries) {
 		queries.forEach(this::maybeCallbackBeforeConvertWithQuery);
 	}
 
+	protected <T> T maybeCallbackAfterSave(T entity) {
+
+		if (entityCallbacks != null) {
+			return entityCallbacks.callback(AfterSaveCallback.class, entity);
```
sothawo: We should make sure that the value returned from the callback is not `null`. Although the API has neither the parameter or return value marked as nullable, we might get a null value by some bad implemention of an after-save callback. So I think we should assert this here.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -117,8 +119,13 @@ public void setEntityCallbacks(EntityCallbacks entityCallbacks) {
...
 		Assert.notNull(index, "index must not be null");
 
-		index(getIndexQuery(entity), index);
-		return entity;
+		IndexQuery query = getIndexQuery(entity);
+		index(query, index);
+
+		// suppressing because it's either entity itself or something of a correct type returned by an entity callback
+		@SuppressWarnings("unchecked")
+		T castResult = (T) query.getObject();
```
sothawo: This must not be null, see my comment for the `maybeCallbackAfterSave(T entity)` method.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -151,7 +158,10 @@ public void setEntityCallbacks(EntityCallbacks entityCallbacks) {
 			});
 		}
 
-		return entities;
+		return indexQueries.stream()
+				.map(IndexQuery::getObject)
```
sothawo: This must not be null, see my comment for the `maybeCallbackAfterSave(T entity)` method.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTransportTemplateUnitTests.java
```diff
@@ -0,0 +1,288 @@
...
+import org.springframework.data.mapping.callback.EntityCallbacks;
+import org.springframework.lang.Nullable;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * @author Roman Puchkovskiy
+ */
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+public class ElasticsearchTransportTemplateUnitTests {
```
sothawo: This should be `ElasticsearchTransportTemplateCallbackTests`, see my comment for the `ElasticsearchRestTemplateUnitTests` class.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplateUnitTests.java
```diff
@@ -0,0 +1,276 @@
...
+import org.springframework.data.mapping.callback.EntityCallbacks;
+import org.springframework.lang.Nullable;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * @author Roman Puchkovskiy
+ */
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+public class ElasticsearchRestTemplateUnitTests {
```
sothawo: I think this should be better named `ElasticsearchRestTemplateCallbackTests`, otherwise it will be confusing with the already existing `ElasticsearchTransportTemplateTests`, and this class is especially for the callbacks. For the reactive part you have named it according to this pattern as well.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
rpuch: Looking at the code of `SimpleEntityCallbackInvoker`, it looke like it makes sure that the value returned by a callback is non-null:

```
				throw new IllegalArgumentException(
						String.format("Callback invocation on %s returned null value for %s", callback.getClass(), entity));
```
Also, javadoc for `EntityCallbackInvoker#invokeCallback()` says that the returned value is 'never null'.
So even a callback that returns null will just cause an exception somewhere in EntityCallbacks implementation, and here we will not get a null value.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
rpuch: `query.getObject()` can be null either if `entity` is null (which is impossible, there is an assertion in the beginning of the method), or if `entityCallbacks.callback()` returns null (which is also impossible).
So the `castResult` cannot be null.

Would you still prefer to assert that is not null, just in case?

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
rpuch: Currently, this can only be null if some entity in the original Iterable is null.

Do you still suggest adding a non-null assertion?

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTransportTemplateUnitTests.java
rpuch: Renamed the test suite to `ElasticsearchTransportTemplateCallbackTests`, thanks for the suggestion

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplateUnitTests.java
rpuch: Renamed the suite to `ElasticsearchRestTemplateCallbackTests`, thanks for the suggestion

rpuch:
Thank you for your review, @sothawo !

I've renamed the test suites.

According to the null-checks, please see my comments.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
sothawo: ah, I didn't dig this deep into the code for this PR, sorry (wish this were Kotlin stuff!), so that's ok

sothawo:
thanks a lot, this is merged.

PR #415, opened 2020-03-30
DATAES-773 - Add search-as-you-type field support

alesharik:
This PR adds search-as-you-type field support


alesharik:
I re-implemented it, but now we have confusing `maxShingleSize` property with default value of 3, because we can't put nulls into annotation properties :/

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -148,4 +149,9 @@
 	 * @since 4.0
 	 */
 	double scalingFactor() default 1;
+
+	/**
+	 * @since 4.0
+	 */
+	int maxShingleSize() default 3;
```
sothawo:
```suggestion
	int maxShingleSize() default -1;
```
we should use -1 here to denote a value that is not set (I forgot that we can't use null)

src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
```diff
@@ -117,4 +118,9 @@
 	 * @since 4.0
 	 */
 	double scalingFactor() default 1;
+
+	/**
+	 * @since 4.0
+	 */
+	int maxShingleSize() default 3;
```
sothawo:
```suggestion
	int maxShingleSize() default -1;
```

same as above

src/main/java/org/springframework/data/elasticsearch/core/index/MappingParameters.java
```diff
@@ -88,6 +90,7 @@
 	private Similarity similarity = Similarity.Default;
 	private TermVector termVector = TermVector.none;
 	private double scalingFactor = 1.0;
+	private int maxShingleSize;
```
sothawo:
```suggestion
	private Integer maxShingleSize = null;
```

here we can have a nullable field

src/main/java/org/springframework/data/elasticsearch/core/index/MappingParameters.java
```diff
@@ -136,6 +139,9 @@ private MappingParameters(Field field) {
 		similarity = field.similarity();
 		termVector = field.termVector();
 		scalingFactor = field.scalingFactor();
+		maxShingleSize = field.maxShingleSize();
```
sothawo:
```suggestion
		maxShingleSize = field.maxShingleSize() >= 0 ? field.maxShingleSize : null;
```
The check in the Assert should account for a not set default value of null

src/main/java/org/springframework/data/elasticsearch/core/index/MappingParameters.java
```diff
@@ -165,6 +171,9 @@ private MappingParameters(InnerField field) {
...
+		Assert.isTrue(type != FieldType.Search_As_You_Type || (maxShingleSize >= 2 && maxShingleSize <= 4),
+				"maxShingleSize must be in inclusive range from 2 to 4 for field type search_as_you_type");
 	}
```
sothawo: see my comment in the other constructor

src/main/java/org/springframework/data/elasticsearch/core/index/MappingParameters.java
```diff
@@ -269,5 +278,9 @@ public void writeTypeAndParametersTo(XContentBuilder builder) throws IOException
 		if (type == FieldType.Scaled_Float) {
 			builder.field(FIELD_PARAM_SCALING_FACTOR, scalingFactor);
 		}
+
+		if (maxShingleSize != 3) {
```
sothawo:
```suggestion
		if (maxShingleSize != null) {
```

sothawo:
Your PR has been merged, thank you again for this. And sorry, that I didn't find the time to merge before the release of 4.0.0.RC1 this morning.

PR #427, opened 2020-04-12
DATAES-786 - Move the creation of SearchHit(s) from ElasticsearchConv…

rpuch:
…erter closer to ElasticsearchTemplate.

The idea is to facilitate entity callbacks invocation (see DATAES-772).
The corresponding code is moved from ElasticsearchConverter closer to ElasticsearchTemplate implementations. It is basically separated into two parts: the 'entity reading' part (that materializes entities from Document and SearchDocument instances) and SearchHit/SearchHits/SearchScrollHits construction part. These parts are separated because the 'entity reading' will later include callback invocations, so it will use different interfaces (reactive/non-reactive) according to the context, while SearchHit/etc construction stays the same.
'entity reading' is moved to DocumentCallbacks that are introduced in this PR.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2013-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.springframework.data.elasticsearch.core;
 
 import java.util.ArrayList;
@@ -13,14 +28,17 @@ import org.elasticsearch.action.bulk.BulkItemResponse;
 import org.elasticsearch.action.bulk.BulkResponse;
 import org.elasticsearch.action.search.MultiSearchRequest;
 import org.elasticsearch.action.search.MultiSearchResponse;
+import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.index.query.MoreLikeThisQueryBuilder;
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
+import org.springframework.data.convert.EntityReader;
 import org.springframework.data.elasticsearch.ElasticsearchException;
 import org.springframework.data.elasticsearch.core.convert.ElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.convert.MappingElasticsearchConverter;
+import org.springframework.data.elasticsearch.core.document.Document;
 import org.springframework.data.elasticsearch.core.document.SearchDocumentResponse;
 import org.springframework.data.elasticsearch.core.event.AfterSaveCallback;
 import org.springframework.data.elasticsearch.core.event.BeforeConvertCallback;
@@ -38,6 +56,7 @@ import org.springframework.data.elasticsearch.support.VersionInfo;
 import org.springframework.data.mapping.callback.EntityCallbacks;
 import org.springframework.data.util.CloseableIterator;
 import org.springframework.data.util.Streamable;
+import org.springframework.lang.NonNull;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
@@ -235,7 +254,7 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 
 		return StreamQueries.streamResults( //
 				searchScrollStart(scrollTimeInMillis, query, clazz, index), //
-				scrollId -> searchScrollContinue(scrollId, scrollTimeInMillis, clazz), //
+				scrollId -> searchScrollContinue(scrollId, scrollTimeInMillis, clazz, index), //
 				this::searchScrollClear);
 	}
 
@@ -262,10 +281,11 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 
 		MultiSearchResponse.Item[] items = getMultiSearchResult(request);
 
+		SearchDocumentResponseCallback<SearchHits<T>> callback = new ReadSearchDocumentResponseCallback<>(clazz, index);
 		List<SearchHits<T>> res = new ArrayList<>(queries.size());
 		int c = 0;
 		for (Query query : queries) {
-			res.add(elasticsearchConverter.read(clazz, SearchDocumentResponse.from(items[c++].getResponse())));
+			res.add(callback.doWith(SearchDocumentResponse.from(items[c++].getResponse())));
 		}
 		return res;
 	}
@@ -285,7 +305,13 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 		int c = 0;
 		Iterator<Class<?>> it1 = classes.iterator();
 		for (Query query : queries) {
-			res.add(elasticsearchConverter.read(it1.next(), SearchDocumentResponse.from(items[c++].getResponse())));
+			Class entityClass = it1.next();
+
+			SearchDocumentResponseCallback<SearchHits<?>> callback = new ReadSearchDocumentResponseCallback<>(
+					entityClass, index);
+
+			SearchResponse response = items[c++].getResponse();
+			res.add(callback.doWith(SearchDocumentResponse.from(response)));
 		}
 		return res;
 	}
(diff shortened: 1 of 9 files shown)


rpuch:
A couple of highlights:

1. `SearchHit`/etc creation is separated from entity materialization because the entity materialization is going to be 'wrapped' with entity callbacks that may be reactive or non-reactive (depending on the context), and those require different interfaces. At the same time, `SearchHit`/etc creation is the same in both contexts.
2. Naming troubles: I'm not sure that `SearchHitMapping` (and names inside it) are the best possible names. I could not come up with anything better for now. What troubles me here is that what is actually done by the methods is 'materialize a `SearchHit`/`SearchHits`/`SearchScrollHits` from `SearchDocument`/`SearchDocumentResponse` *given entity/entities that are already mapped*. It's too long, so I just named them `mapXXX()`.

That do you think?

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -285,11 +309,22 @@ public long count(Query query, Class<?> clazz) {
...
+			SearchDocumentResponseCallback<SearchHits<?>> callback = readSearchDocumentResponseCallback(
+					entityClass, index);
+
+			SearchResponse response = items[c++].getResponse();
+			res.add(callback.doWith(SearchDocumentResponse.from(response)));
 		}
 		return res;
 	}
 
+	private <T> SearchDocumentResponseCallback<SearchHits<T>> readSearchDocumentResponseCallback(Class<T> entityClass,
```
sothawo, line 323: can we cache these for (entityClass, index)? to prevent new creation?
 Are the callbacks then threadsafe, if we reuse cached instances?

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -1,5 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.springframework.data.elasticsearch.core;
 
+import lombok.NonNull;
```
sothawo, line 18: Please don't use lombok in production code, only in test code.
For nullable or not null annotation use the ones from the `org.springframework.lang` package.

src/main/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchTemplate.java
```diff
@@ -894,4 +895,55 @@ private RuntimeException translateException(Throwable throwable) {
 		return Mono.just(entity);
 	}
 	// endregion
+
+	protected interface DocumentCallback<T> {
```
sothawo, line 899: first thought was that using the same name as in non-reactive code might be not the best, but as these are protected nested classes, this should be no problem.

src/main/java/org/springframework/data/elasticsearch/core/SearchHitMappings.java
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import lombok.NonNull;
```
sothawo, line 18: see my other comment

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
rpuch: I've removed lombok's annotation and added `Assert.notNull()` instead.

src/main/java/org/springframework/data/elasticsearch/core/SearchHitMappings.java
rpuch: I've removed lombok usage here

sothawo:
as for the naming: In other places the classes that create interface implementations or stuff like that are called `...Adapter` (See `DocumentAdapter` to create `Document` implementations ). Perhaps `Search(Scroll)Hit(s)Adapter` might be a solution.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
rpuch: What's the point? `ReadSearchDocumentResponseCallback` is created once per query. A query execution takes orders of magnitide more time than callback object allocation (because it involves network I/O and, probably, disk I/O on the server).
In terms of garbage, one or two objects per query (given that each query may produce dozens, hundreds or more entities) does not seem to be a high price as well.

rpuch:
> as for the naming: In other places the classes that create interface implementations or stuff like that are called `...Adapter` (See `DocumentAdapter` to create `Document` implementations ). Perhaps `Search(Scroll)Hit(s)Adapter` might be a solution.

Could you please elaborate: what class(es) do you suggest to rename and how exactly? To be honest, I thought that class naming was more or less ok, I had doubts about method naming inside `SearchHitMapping` :)

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
sothawo: forget about that caching idea.

`readSearchDocumentResponseCallback(...)` is a private method that's a one-liner and is only called at one place in the second _multiSearch_ method. I think we could inline this method.

rpuch: The only reason I felt this method was needed was to capture the type variable. Without this, it would require to use even more raw types. Are you ok with this?

src/main/java/org/springframework/data/elasticsearch/core/SearchHitMappings.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * @author Rizwan Idrees
+ * @author Mohsin Husen
+ * @author Christoph Strobl
+ * @author Peter-Josef Meisch
+ * @author Mark Paluch
+ * @author Roman Puchkovskiy
+ * @since 4.0
+ */
+class SearchHitMappings {
```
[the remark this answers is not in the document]
rpuch: To be honest, I usually prefer to reduce the number of method parameters. To have one field more is not a big deal, it's somewhere in the 'data' section of the class and almost never gets in your way when you read the code in methods. But each extra parameter clutters the *active* code itself.

This nested structure is not too elegant, but it is hidden/encapsulated in the class where it is defined, and does not irritate the callers. But as a result, the `mappingContext` is encapsulated and does not need to be passed.

Please let me know if this is convincing or not. If not, I will surely follow your advice :)

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
sothawo: If I inline this I get:

```
for (Query query : queries) {
	Class entityClass = it1.next();

	SearchDocumentResponseCallback<SearchHits<?>> callback = new ReadSearchDocumentResponseCallback<>(entityClass, index);

	SearchResponse response = items[c++].getResponse();
	res.add(callback.doWith(SearchDocumentResponse.from(response)));
}
```
same as in the other `multiSearch` method above this one.

src/main/java/org/springframework/data/elasticsearch/core/SearchHitMappings.java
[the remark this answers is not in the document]
sothawo: I'd rather have a second method parameter than the nested class construct. But you are right, here this would need to be intorduced in all the methods, becuase the calling ones need to have it as well.
Haveing a closer look, But we don't need neither the additional parameter nor the outer class. When removing the outer `SearchHitMappings` and keeping the inner this way:
```
class SearchHitMapping<T> {

	private final MappingContext<? extends ElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty> mappingContext;
	private final Class<T> type;

	static <T> SearchHitMapping<T> mappingFor(Class<T> entityClass,
			MappingContext<? extends ElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty> mappingContext) {
		return new SearchHitMapping<>(entityClass, mappingContext);
	}

	private SearchHitMapping(Class<T> type,
			MappingContext<? extends ElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty> mappingContext) {
		Assert.notNull(type, "type must not be null");
		Assert.notNull(mappingContext, "mappingContext must not be null");

		this.type = type;
		this.mappingContext = mappingContext;
	}

        // all the other methods
}
```
we have one class and we don't need the additional fields in the template classes; there we then have:
```
public SearchHits<T> doWith(SearchDocumentResponse response) {
	List<T> entities = response.getSearchDocuments().stream()
			.map(delegate::doWith)
			.collect(Collectors.toList());
	return SearchHitMapping.mappingFor(type, elasticsearchConverter.getMappingContext()).mapHits(response, entities);
}
```

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
rpuch: Oh, sure, I introduced this method when 2 separate callbacks were created in it with the same type parameter. Now it is not needed, I've inlined it. Thanks.

src/main/java/org/springframework/data/elasticsearch/core/SearchHitMappings.java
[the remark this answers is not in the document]
rpuch: Yes, I like this too, I've applied changes according to your suggestion. Thanks.

rpuch:
It looks like that's it with the requested changes if I did not forget something.

PR #430, opened 2020-04-18
DATAES-567 - Add aggregation support to reactive client.

amordleq:
Related tickets: DATA-695

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
@@ -410,6 +410,12 @@ public class DefaultReactiveElasticsearchClient implements ReactiveElasticsearch
 	 */
 	@Override
 	public Flux<Aggregation> aggregate(HttpHeaders headers, SearchRequest searchRequest) {
+		
+		Assert.notNull(headers, "headers must not be null");
+		Assert.notNull(searchRequest, "searchRequest must not be null");
+		
+		searchRequest.source().size(0);
+		
 		return sendRequest(searchRequest, requestCreator.search(), SearchResponse.class, headers) //
 				.map(SearchResponse::getAggregations) //
 				.flatMap(Flux::fromIterable);
--- a/src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
@@ -94,6 +94,7 @@ import org.elasticsearch.rest.RestStatus;
 import org.elasticsearch.search.Scroll;
 import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.search.SearchHits;
+import org.elasticsearch.search.aggregations.Aggregation;
 import org.reactivestreams.Publisher;
 import org.springframework.data.elasticsearch.client.ClientConfiguration;
 import org.springframework.data.elasticsearch.client.ClientLogger;
@@ -101,6 +102,7 @@ import org.springframework.data.elasticsearch.client.ElasticsearchHost;
 import org.springframework.data.elasticsearch.client.NoReachableHostException;
 import org.springframework.data.elasticsearch.client.reactive.HostProvider.Verification;
 import org.springframework.data.elasticsearch.client.reactive.ReactiveElasticsearchClient.Indices;
+import org.springframework.data.elasticsearch.client.util.NamedXContents;
 import org.springframework.data.util.Lazy;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
@@ -128,6 +130,7 @@ import org.springframework.web.reactive.function.client.WebClient.RequestBodySpe
  * @author Huw Ayling-Miller
  * @author Henrique Amaral
  * @author Roman Puchkovskiy
+ * @author Russell Parry
  * @since 3.2
  * @see ClientConfiguration
  * @see ReactiveRestClients
@@ -398,6 +401,17 @@ public class DefaultReactiveElasticsearchClient implements ReactiveElasticsearch
 
 		return sendRequest(searchRequest, requestCreator.search(), SearchResponse.class, headers) //
 				.map(SearchResponse::getHits) //
+				.flatMap(Flux::fromIterable);
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * @see org.springframework.data.elasticsearch.client.reactive.ReactiveElasticsearchClient#aggregate(org.springframework.http.HttpHeaders, org.elasticsearch.action.search.SearchRequest)
+	 */
+	@Override
+	public Flux<Aggregation> aggregate(HttpHeaders headers, SearchRequest searchRequest) {
+		return sendRequest(searchRequest, requestCreator.search(), SearchResponse.class, headers) //
+				.map(SearchResponse::getAggregations) //
 				.flatMap(Flux::fromIterable);
 	}
 
@@ -751,10 +765,9 @@ public class DefaultReactiveElasticsearchClient implements ReactiveElasticsearch
 	}
 
 	private static XContentParser createParser(String mediaType, String content) throws IOException {
-
 		return XContentType.fromMediaTypeOrFormat(mediaType) //
 				.xContent() //
-				.createParser(NamedXContentRegistry.EMPTY, DeprecationHandler.THROW_UNSUPPORTED_OPERATION, content);
+				.createParser(new NamedXContentRegistry(NamedXContents.getDefaultNamedXContents()), DeprecationHandler.THROW_UNSUPPORTED_OPERATION, content);
 	}
 
 	private static <T> Publisher<? extends T> handleServerError(Request request, ClientResponse response) {
--- a/src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClient.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClient.java
@@ -16,6 +16,7 @@
 package org.springframework.data.elasticsearch.client.reactive;
 
 import org.elasticsearch.search.aggregations.Aggregation;
+import org.springframework.util.Assert;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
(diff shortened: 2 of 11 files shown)


src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClient.java
```diff
@@ -411,6 +412,41 @@
 	 */
 	Flux<SearchHit> search(HttpHeaders headers, SearchRequest searchRequest);
 
+	/**
+	 * Execute the given {@link SearchRequest} with aggregations against the {@literal search} API.
+	 *
+	 * @param consumer never {@literal null}.
+	 * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html">Search API on
+	 *      elastic.co</a>
+	 * @return the {@link Flux} emitting {@link Aggregation} one by one.
```
sothawo, line 421:
```suggestion
	 * @return the {@link Flux} emitting {@link Aggregation} one by one.
	 * @since 4.0
```

src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClient.java
```diff
@@ -411,6 +412,41 @@
...
+		consumer.accept(request);
+		return aggregate(request);
+	}
+	/**
+	 * Execute the given {@link SearchRequest} with aggregations against the {@literal search} API.
+	 *
+	 * @param searchRequest must not be {@literal null}.
+	 * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html">Search API on
+	 *      elastic.co</a>
+	 * @return the {@link Flux} emitting {@link Aggregation} one by one.
```
sothawo, line 435:
```suggestion
	 * @return the {@link Flux} emitting {@link Aggregation} one by one.
	 * @since 4.0
```

src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClient.java
```diff
@@ -411,6 +412,41 @@
...
+	default Flux<Aggregation> aggregate(SearchRequest searchRequest) { return aggregate(HttpHeaders.EMPTY, searchRequest); }
+
+	/**
+	 * Execute the given {@link SearchRequest} with aggregations against the {@literal search} API.
+	 *
+	 * @param headers Use {@link HttpHeaders} to provide eg. authentication data. Must not be {@literal null}.
+	 * @param searchRequest must not be {@literal null}.
+	 * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html">Search API on
+	 *      elastic.co</a>
+	 * @return the {@link Flux} emitting {@link Aggregation} one by one.
```
sothawo, line 446:
```suggestion
	 * @return the {@link Flux} emitting {@link Aggregation} one by one.
 	 * @since 4.0
```

src/main/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClient.java
```diff
@@ -411,6 +412,41 @@
...
+	/**
+	 * Execute the given {@link SearchRequest} with aggregations against the {@literal search} API.
+	 *
+	 * @param consumer never {@literal null}.
+	 * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html">Search API on
+	 *      elastic.co</a>
+	 * @return the {@link Flux} emitting {@link Aggregation} one by one.
+	 */
+	default Flux<Aggregation> aggregate(Consumer<SearchRequest> consumer) {
+
```
sothawo, line 424:
```suggestion
    Assert.notNull(consumer, "consumer must not be null"):
```

Parameters that are not annotated as `@Nullable` must be checked.

src/main/java/org/springframework/data/elasticsearch/client/reactive/DefaultReactiveElasticsearchClient.java
```diff
@@ -401,6 +404,17 @@ public Indices indices() {
 				.flatMap(Flux::fromIterable);
 	}
 
+	/*
+	 * (non-Javadoc)
+	 * @see org.springframework.data.elasticsearch.client.reactive.ReactiveElasticsearchClient#aggregate(org.springframework.http.HttpHeaders, org.elasticsearch.action.search.SearchRequest)
+	 */
+	@Override
+	public Flux<Aggregation> aggregate(HttpHeaders headers, SearchRequest searchRequest) {
+		return sendRequest(searchRequest, requestCreator.search(), SearchResponse.class, headers) //
```
sothawo, line 413:
```suggestion
		
		Assert.notNull(headers, "headers must not be null");
		Assert.notNull(searchRequest, "searchRequest must not be null");
		
		searchRequest.source().size(0);
		
		return sendRequest(searchRequest, requestCreator.search(), SearchResponse.class, headers) //
```

we should check non-nullable parameters and as we don't need the source in the response, we should set it to zero

src/test/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClientUnitTests.java
```diff
@@ -577,6 +582,83 @@ public void searchShouldReturnEmptyFluxIfNothingFound() {
 				.verifyComplete();
 	}
 
+	// --> AGGREGATE
```
sothawo, line 585: when we do not request the source, I think we don't need the test cases checking for the source (and the result json files)

amordleq:
Please bear with me a bit, as I'm new to contributing like this.  For the suggestions above that were simple changes, I used github's "Commit Changes".  The last comment obviously requires more than that and I can work on that part.
But the changes I've accepted have now been committed as individual commits and don't follow the commit message guidelines.  Have I screwed something up at this point?

amordleq:
For the `ReactiveElasticsearchTemplate` part, I've actually only been using the client and not the full template with mapping and all of that.  I can definitely look into adding that portion for completeness though.  Should that be a commit against this same branch/pull request in order to make this "complete", or should that be part of something separate somehow?

sothawo:
> For the `ReactiveElasticsearchTemplate` part, I've actually only been using the client and not the full template with mapping and all of that. I can definitely look into adding that portion for completeness though. Should that be a commit against this same branch/pull request in order to make this "complete", or should that be part of something separate somehow?

This PR. There's no mapping involved here, as we don't return any entity data. So it's just ttaking a `Query`, calling the client implementation and returning the aggregates flux.

src/test/java/org/springframework/data/elasticsearch/client/reactive/ReactiveElasticsearchClientUnitTests.java
amordleq: Committed a change for this.

amordleq:
I apparently shouldn't have used github's change feature because I also didn't notice there was a minor syntax typo (which then caused the build to fail).  So I fixed that along with my latest normal commit but now the history on this branch is a little bit of a mess.  Let me know if you want me to do some sort of history rewrite/force push or how else I should clean things up.  In the meantime I'll start looking into the template side.

sothawo:
> I apparently shouldn't have used github's change feature because I also didn't notice there was a minor syntax typo (which then caused the build to fail). So I fixed that along with my latest normal commit but now the history on this branch is a little bit of a mess. Let me know if you want me to do some sort of history rewrite/force push or how else I should clean things up. In the meantime I'll start looking into the template side.

squash it together when you're done with the rest of the changes (I'll do a squash and merge anyway)

amordleq:
I think I have the template support added if I've understood what you were saying.  I'm treating aggregation more like `count` and less like `search`, where no entity mapping of return types needs to happen.  This is a little different than the only other Spring Data experience I've had with spring-data-mongo.  There, you can map aggregation results into entities and have those entities returned from an aggregation call.  But that's a totally different sort of beast than Elasticsearch aggregations and I'm not sure what entity mapping in the ES case would even mean.  There would at least be some other discussion required.  But it sounds like that's not what you were expecting anyway, so I think this commit should do what you were looking for.
Holding off on squashing the commits in case there are more iterations here or I've misunderstood what you're looking for in the template.  Let me know if everything looks good and you want me to do that (or if you just want to squash everything on merge).

sothawo:
Thanks, I have squashed, merged and polished this PR.
As for returning the aggregations: In Spring Data Elasticsearch, the entity is the document stored in the index, and when returning entities we always talk about these objects. This is different from Spring Data Mongodb. 
I'd like to have our own objects to return aggregation values, but this is something to be implemented in the futer.

Dav1dde:
@sothawo @amordleq I made the original Issue in Jira, from what I can see you can now execute aggregations but you cannot execute a search and aggregation at the same time, you would need to separate requests for this.

But Search + Aggregate is the primary usecase of any facetted search e.g. for a shop (Amazon etc.). The aggregations give you additional filters and search/query the items to display.

I was hoping for an API that allows to do both with one query against elasticsearch (like the Sync ElasticSearchOperations does)

amordleq:
@Dav1dde when I originally contacted @sothawo about contributing a PR he asked about my proposed approach because, I think, it's not clear what the reactive API should be if you want _both_ the search hits and the aggregation results.  What would that be a `Flux` of?  Or maybe it should be a `Mono` of a `SearchResponse`?  At any rate, I made the case that it would be useful to at least be able to get the aggregations without the search hits, because there are lots of use cases where you only care about those (which, indeed, is my use case).  And that's a much easier API to reason about because it returns a single `Flux<Aggregation>`.  

Sounds like that doesn't fully handle your use case though.  Do you have thoughts on what you'd expect a reactive API returning both search hits and aggregation results would look like?  I might have some time to keep contributing to this if we could all agree on what a reasonable API would look like here.

Dav1dde:
When I initially created the bug report I needed a solution so I hacked myself an interface together:

```java
public interface ExtendedElasticOperations {
    ReactiveElasticsearchOperations getElasticsearchOperations();

    Mono<SearchResponse> query(SearchRequest query);
}
```

This somewhat follows the [undocumented method in `ElasticSearchOperations`](https://docs.spring.io/spring-data/elasticsearch/docs/current/api/org/springframework/data/elasticsearch/core/ElasticsearchOperations.html#query-org.springframework.data.elasticsearch.core.query.SearchQuery-org.springframework.data.elasticsearch.core.ResultsExtractor-), the result extractor doesn't really make sense in a Mono/Flux environment imo.

This is has been serving me really well in a smaller project with an analysis usecase and a shop usecase.

I think having a more "advanced" interface that gives you the `SearchResponse` is a good thing anyways.

PS: Thanks for your work on this, unfortunately I did not have the time to polish my implementation and contribute it. Even without the method I need this is going to greatly reduce the code that is required to setup my parser.

indrgun:
@sothawo Can this PR be back-ported to 3.2.x branch so it is available for 3.2.x.RELEASE?

Thank you
-Indra

sothawo:
@indrgun not without really major works. This would mean to apply changes dating back to end of March as well, because internally the reactive code was restructured since then.

version 4.0 should be GA within the next one or two weeks, so you could switch to version 4 then, but which would mean to update to Elasticsearch 7.

PR #431, opened 2020-04-19
DATAES-785 - Various entity callbacks implementation improvements.

rpuch:
* Add IndexCoordinates parameter to [Reactive]AfterSaveCallback.
* Add IndexCoordinates parameter to [Reactive]BeforeConvertCallback.
* Make sure before-convert callbacks are invoked in reactive save-all case.

The third task (before-convert callbacks in reactive save-all case) made me make the code rather involved. I tried to reduce its complexity by introducing a helper class, `Entities`.


sothawo:
will have a look at this tomorrow

PR #467, opened 2020-05-23
DATAES-840 - Introduce IndexCoordinateResolver.

sothawo:
This PR introduces the `IndexCordinateResolver` class that is used to resolve index names that are defined with SpEL expressions. Up to now this was handled in the `SimpleElasticsearchPersistentEntity` class, but we will need this functionality for functions that are not related to entities as well (alias management, index templates, support for rolling indices).

The aim is later to support a definition like the following:
 ```
@Document(indexName="log-all", 
    writeIndexName="log-#{T(java.time.LocalDate).now().toString()}")
```
with _log-all_ being an alias (or just use _log-*_ as indexName).

So the following has changed:
* all `*Operations` implementations now use the `RequestFactory` to have the requests created that are sent to Elasticsearch
* the `RequestFactory` class takes care of resolving the index names for the requests to build
* new `IndexCoordinateResolver` class that lives within the `MappingElasticsearchConverter` and is used by the `RequestFactory`.
* the code that was used up to now for evaluating SpEL expressions in the `SimpleElasticsearchPersistentEntity` is removed.

The changes in `RequestFactory` look more than what was actually done. Basically there were some functions added, but I rearranged and grouped them. I wait for the day that the `TransportClient` goes away, then this file will shrink to half it's size.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
@@ -32,6 +32,7 @@ import org.elasticsearch.action.search.MultiSearchResponse;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.index.query.MoreLikeThisQueryBuilder;
+import org.elasticsearch.search.suggest.SuggestBuilder;
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
@@ -48,6 +49,7 @@ import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersiste
 import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentProperty;
 import org.springframework.data.elasticsearch.core.mapping.IndexCoordinates;
 import org.springframework.data.elasticsearch.core.mapping.SimpleElasticsearchMappingContext;
+import org.springframework.data.elasticsearch.core.query.BulkOptions;
 import org.springframework.data.elasticsearch.core.query.GetQuery;
 import org.springframework.data.elasticsearch.core.query.IndexQuery;
 import org.springframework.data.elasticsearch.core.query.IndexQueryBuilder;
@@ -55,6 +57,7 @@ import org.springframework.data.elasticsearch.core.query.MoreLikeThisQuery;
 import org.springframework.data.elasticsearch.core.query.NativeSearchQueryBuilder;
 import org.springframework.data.elasticsearch.core.query.Query;
 import org.springframework.data.elasticsearch.core.query.SeqNoPrimaryTerm;
+import org.springframework.data.elasticsearch.core.query.UpdateQuery;
 import org.springframework.data.elasticsearch.support.VersionInfo;
 import org.springframework.data.mapping.callback.EntityCallbacks;
 import org.springframework.data.util.CloseableIterator;
@@ -200,6 +203,11 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	}
 
 	@Override
+	public <T> List<T> multiGet(Query query, Class<T> clazz) {
+		return multiGet(query, clazz, getIndexCoordinatesFor(clazz));
+	}
+
+	@Override
 	@Nullable
 	public <T> T queryForObject(GetQuery query, Class<T> clazz) {
 		return get(query.getId(), clazz, getIndexCoordinatesFor(clazz));
@@ -227,6 +235,11 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	}
 
 	@Override
+	public void delete(Query query, Class<?> clazz) {
+		delete(query, getIndexCoordinatesFor(clazz));
+	}
+
+	@Override
 	public String delete(Object entity) {
 		return delete(entity, getIndexCoordinatesFor(entity.getClass()));
 	}
@@ -235,6 +248,22 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	public String delete(Object entity, IndexCoordinates index) {
 		return this.delete(getEntityId(entity), index);
 	}
+
+	@Override
+	public List<String> bulkIndex(List<IndexQuery> queries, Class<?> clazz) {
+		return bulkIndex(queries, getIndexCoordinatesFor(clazz));
+	}
+
+	@Override
+	public List<String> bulkIndex(List<IndexQuery> queries, BulkOptions bulkOptions, Class<?> clazz) {
+		return bulkIndex(queries, bulkOptions, getIndexCoordinatesFor(clazz));
+	}
+
+	@Override
+	public void bulkUpdate(List<UpdateQuery> queries, Class<?> clazz) {
+		bulkUpdate(queries, getIndexCoordinatesFor(clazz));
+	}
+
 	// endregion
 
 	// region SearchOperations
@@ -283,6 +312,11 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	}
 
 	@Override
+	public <T> List<SearchHits<T>> multiSearch(List<? extends Query> queries, Class<T> clazz) {
+		return multiSearch(queries, clazz, getIndexCoordinatesFor(clazz));
+	}
+
+	@Override
 	public <T> List<SearchHits<T>> multiSearch(List<? extends Query> queries, Class<T> clazz, IndexCoordinates index) {
 		MultiSearchRequest request = new MultiSearchRequest();
 		for (Query query : queries) {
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: src/main/java/org/springframework/data/elasticsearch/core/mapping/IndexCoordinatesResolver.java [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * A class to resolve index names (or generally Strings) by using an ApplicationContext aware SpELExpressionParser. The
+ * index names to resolve must be SpEL template expressions.
+ *
+ * @author Peter-Josef Meisch
+ */
+public class IndexCoordinatesResolver implements ApplicationContextAware {
+
+	private final StandardEvaluationContext context;
```
mp911de, line 39: SpEL should make use of Spring Data's `EvaluationContext` resolution which is available from within `BasicPersistentEntity`. What's the need for introducing a wrapper for SpEL evaluation?

In other modules, we do not leak outside of the entity/property the fact that a SpEL expression is used for the collection/table name.

sothawo: We do not always have an entity for example in the IndexOperations, for example when `IndexOperations ElasticsearchOperations.indexOps(IndexCoordinates index)` is used.

BTW, I didn't know that there is a EvaluationContext available, the code that I use for the SpEL evaluation was basically moved from  `SimpleElasticsearchPersistenEntity`

sothawo: I'd like to reuse the `EvaluationContextProvider` from `MappingContext`, but it is private

mp911de:
Thanks for the detail. IndexCoordinates should be always resolved otherwise we always need to introduce SpEL awareness to components that touch this type. We should carry this fact around, instead, we should obtain the index coordinates from a persistent entity. That way, we can keep SpEL resolution inside the PersistentEntity and we don’t need to touch that much code.

sothawo:
There is no entity in the update method. Nor in createAlias. deleteByQuery has no entities. In Elasticsearch it is quite valid to have request to work across multiple indices that don't need to have the same type. deleteIndex is another one that can delete multiple indices in one call.

Currently it is possible to get the IndexCoordinates from a PersistentEntity, and it would be possible to do the SpEL evaluation on getting this. That is what we had before this PR - although this did not use the EvaluationContext from the entity.
The reason I moved this into the RequestFactory was to be able to do the evaluation when building the requests and to be able to do this for all requests that we send to Elasticsearch.

If we tie this to the entity we need to deprecate all the methods that don't need an entity class up to now. 

The PersistentEntity gets a EvaluationContextProvider from the MappingContext, I would reuse that if it were accessible from SimpleElasticsearchMappingContext. 

For me, every entity has IndexCoordinates, right. But not the other way round.

mp911de:
Typically, you have either a type, IndexCoordinates or both. Probably I’m missing something. Types without an entity context receive their dynamic through IndexCoordinates but are not able to map queries against a type. 

I did not understand the part of building requests for methods without an entity type. What do you think of jumping onto a call this week so we can continue the discussion in a more direct way?

sothawo:
`IndexCoordinatesResolver` now is nested private in the entity. SpEL resolution is only done in the entity. I added new methods to the Operations so that we have methods with an entity(class) and ones with the `IndexCoordinates`.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchMappingContext.java
```diff
@@ -15,46 +15,30 @@
...
  * SimpleElasticsearchMappingContext
  *
  * @author Rizwan Idrees
  * @author Mohsin Husen
  * @author Mark Paluch
+ * @author Peter-Josef Meisch
  */
 public class SimpleElasticsearchMappingContext
-		extends AbstractMappingContext<SimpleElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty>
-		implements ApplicationContextAware {
```
mp911de, line 36: Un-implementing an interface is a breaking change.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -102,26 +94,13 @@ public SimpleElasticsearchPersistentEntity(TypeInformation<T> typeInformation) {
 		}
 	}
 
-	@Override
-	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
-		context.addPropertyAccessor(new BeanFactoryAccessor());
```
mp911de, line 107: Removing access to the ApplicationContext removes functionality. What’s the motivation for this change?

sothawo: The only purpose of implementing `ApplicationContextAware` was to set up the EvaluationContext for the SpEL resolving of index names. But with this change, the context from the base class is used, so the setup is not needed anymore. Keeping implementing the interface would lead to a noop implementation of `setApplicationContext`

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
```diff
@@ -86,6 +86,20 @@
 	 */
 	<T> Mono<T> save(T entity, IndexCoordinates index);
 
+	/**
+	 * Index entities under the given {@literal type} in the index extracted from entity metadata.
+	 *
+	 * @param entities must not be {@literal null}.
+	 * @param clazz the entity class
```
mp911de, line 93: the javadoc should reflect that the class parameter is used to determine the actual index.

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
```diff
@@ -102,6 +116,16 @@
 		return saveAll(Mono.just(entityList), index);
 	}
 
+	/**
+	 * Index entities under the given {@literal type} in the index extracted from entity metadata.
+	 *
+	 * @param entities must not be {@literal null}.
+	 * @param clazz the entity class
```
mp911de, line 123: See above.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -66,6 +71,15 @@ public SimpleElasticsearchPersistentProperty(Property property,
 
 		this.annotatedFieldName = getAnnotatedFieldName();
 		this.isId = super.isIdProperty() || SUPPORTED_ID_PROPERTY_NAMES.contains(getFieldName());
+
+		// @deprecated since 4.1
+		@Deprecated
```
mp911de, line 76: No need fort inline annotations.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -294,4 +273,55 @@ public boolean hasSeqNoPrimaryTermProperty() {
 	public ElasticsearchPersistentProperty getSeqNoPrimaryTermProperty() {
 		return seqNoPrimaryTermProperty;
 	}
+
+	private static class IndexCoordinatesResolver {
+
+		private final SpelExpressionParser parser;
```
mp911de, line 279: Let's keep the parser as constant so we don't need to create a new instance for each entity. In general, the design could be trimmed down to keeping the parser and a potential `Expression` object (see [`BasicMongoPersistentEntity`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/BasicMongoPersistentEntity.java#L59-L65) for further reference).

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -102,26 +94,13 @@ public SimpleElasticsearchPersistentEntity(TypeInformation<T> typeInformation) {
...
-		}
-
-		return getTypeInformation().getType().getSimpleName();
+		return indexName != null ? indexName : getTypeInformation().getType().getSimpleName();
 	}
 
 	@Override
 	public IndexCoordinates getIndexCoordinates() {
-		return IndexCoordinates.of(getIndexName());
+		return indexCoordinatesResolver.resolve(getEvaluationContext(null), IndexCoordinates.of(getIndexName()));
```
mp911de, line 103: How about simplifying this method back to `IndexCoordinates.of(getIndexName()))` and keeping the evaluation within `getIndexName()`?

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
sothawo: I moved the IndexCoordinateResolver stuff into SimpleElasticsearchPersistentEntity.  But we need more than one expression when we later introduce the writeIndexname in addition to the current one. That's also the reason I kept the resolve in `getIndexCoordinates()` and did not move it into `getIndexName`-- we'd need to either change that later or have two calls to `resolve`

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
sothawo: removed the `@` this comment is just a documentation as hint for later that we know since when we deprecated it

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
sothawo: see above

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/ReactiveDocumentOperations.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -300,6 +334,33 @@ public long count(Query query, Class<?> clazz) {
 		return res;
 	}
 
+	@Override
+	public List<SearchHits<?>> multiSearch(List<? extends Query> queries, List<Class<?>> classes) {
+
+		MultiSearchRequest request = new MultiSearchRequest();
```
mp911de, line 340: Nit: It would make sense to check that `queries.size() == classes.size()`

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
mp911de: I see. Without access to `ApplicationContext`, SpEL expressions are no longer able to reference beans from the context. We've seen some examples where table names/collection names referenced config classes that contributed parts of the table/collection name.

sothawo:
>Without access to ApplicationContext, SpEL expressions are no longer able to reference beans from the context. We've seen some examples where table names/collection names referenced config classes that contributed parts of the table/collection name.

that's not a problem here, because the context that `org.springframework.data.mapping.model.BasicPersistentEntity#getEvaluationContext(Object rootObject)` returns is able to resolve beans.

PR #469, opened 2020-05-29
DATAES-847 Add DateFormat values for epoch_millis and epoch_second; adopt week_year name changes

timtebeek:
- [x] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAES-847).

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/DateFormat.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/DateFormat.java
@@ -17,6 +17,7 @@ package org.springframework.data.elasticsearch.annotations;
 
 /**
  * @author Jakub Vavrik
+ * @author Tim te Beek
  *         Values based on reference doc - https://www.elastic.co/guide/reference/mapping/date-format/
  */
 public enum DateFormat {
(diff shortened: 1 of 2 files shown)


timtebeek:
There's a few more oddities when comparing the fields in DateFormat with the fields listed at ES:
https://github.com/timtebeek/spring-data-elasticsearch/blob/master/src/main/java/org/springframework/data/elasticsearch/annotations/DateFormat.java
https://www.elastic.co/guide/reference/mapping/date-format
```
official elasticsearch docs				      |	DateFormat class
week_date_time_no_millis				      |	weekDateTimeNoMillis
weekyear						      |	week_year
weekyear_week						      |	weekyearWeek
weekyear_week_day					      |	weekyearWeekDay
```

timtebeek:
I'm guessing we'll want to switch the offending constants over to the official references again, as I'm not sure the current camelCased values even work. But also don't want to introduce a potentially breaking change if they do still work, so I'm leaving that up the the maintainers to decide whether we want to change those additional fields.

timtebeek:
It seems the old values originate with any pre ES2:
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/mapping-date-format.html
Starting with 2.0 the fields seem to have been renamed:
https://www.elastic.co/guide/en/elasticsearch/reference/2.0/mapping-date-format.html
As such it would seem the current values wouldn't work and fixing them here would be sensible.

timtebeek:
Awesome; thanks for the quick review & merged; surprised how easy both the fix was and the process to get it approved! :)

sothawo:
You're welcome! (and I could try the new integrated PR feature of IntelliJ IDEA 2020.2 EAP)

PR #482, opened 2020-06-16
DATAES-865 fix NPE at MappingElasticsearchConverter.java

Been24:
when Object value of writeProperty() as ClassTypeInformation.MAP
an error occurred caused by NPE at writeMapValue().
for example:
  private Object content;(the actual type of content is Map<String,Object>)


sothawo:
Thanks you for finding and fixing this. The next time please make sure that you can check off the rest of the points (writing a test, adding your name etc.)

PR #485, opened 2020-06-23
[DATAES-433] Support join datatype

sbcd90:
This PR adds support for join datatype in spring-data-elasticsearch.
The test case for Join is present here:

https://github.com/sbcd90/spring-data-elasticsearch/blob/es_join/src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java#L3028-#L3087

- [ Yes] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ Yes] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAES-433).
- [ Yes] You submit test cases (unit or integration tests) that back your changes.

Kindly review.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
@@ -39,6 +39,8 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.data.convert.EntityReader;
 import org.springframework.data.elasticsearch.BulkFailureException;
+import org.springframework.data.elasticsearch.core.join.JoinField;
+import org.springframework.data.elasticsearch.annotations.JoinTypeRelations;
 import org.springframework.data.elasticsearch.core.convert.ElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.convert.MappingElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.document.Document;
@@ -533,6 +535,22 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	}
 
 	@Nullable
+	private String getEntityRouting(Object entity) {
+		ElasticsearchPersistentEntity<?> persistentEntity = getRequiredPersistentEntity(entity.getClass());
+		ElasticsearchPersistentProperty joinProperty = persistentEntity.getJoinFieldProperty();
+
+		if (joinProperty != null) {
+			Object joinField = persistentEntity.getPropertyAccessor(entity).getProperty(joinProperty);
+			if (joinField != null && JoinField.class.isAssignableFrom(joinField.getClass())
+				&& ((JoinField<?>) joinField).getParent() != null) {
+				return elasticsearchConverter.convertId(((JoinField<?>) joinField).getParent());
+			}
+		}
+
+		return null;
+	}
+
+	@Nullable
 	private Long getEntityVersion(Object entity) {
 		ElasticsearchPersistentEntity<?> persistentEntity = getRequiredPersistentEntity(entity.getClass());
 		ElasticsearchPersistentProperty versionProperty = persistentEntity.getVersionProperty();
@@ -580,6 +598,11 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 		} else {
 			// version cannot be used together with seq_no and primary_term
 			builder.withVersion(getEntityVersion(entity));
+		}
+
+		String routing = getEntityRouting(entity);
+		if (routing != null) {
+			builder.withRouting(routing);
 		}
 		return builder.build();
 	}
--- a/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
@@ -743,6 +743,10 @@ class RequestFactory {
 			deleteByQueryRequest.setScroll(TimeValue.timeValueMillis(query.getScrollTime().toMillis()));
 		}
 
+		if (query.getRoute() != null) {
+			deleteByQueryRequest.setRouting(query.getRoute());
+		}
+
 		return deleteByQueryRequest;
 	}
 
@@ -796,6 +800,10 @@ class RequestFactory {
 		if (query.hasScrollTime()) {
 			// noinspection ConstantConditions
 			source.setScroll(TimeValue.timeValueMillis(query.getScrollTime().toMillis()));
+		}
+
+		if (query.getRoute() != null) {
+			source.setRouting(query.getRoute());
 		}
 
 		return requestBuilder;
@@ -888,6 +896,10 @@ class RequestFactory {
 			indexRequest.setIfPrimaryTerm(query.getPrimaryTerm());
 		}
 
+		if (query.getRouting() != null) {
+			indexRequest.routing(query.getRouting());
+		}
+
 		return indexRequest;
 	}
 
@@ -925,6 +937,9 @@ class RequestFactory {
 		}
 		if (query.getPrimaryTerm() != null) {
 			indexRequestBuilder.setIfPrimaryTerm(query.getPrimaryTerm());
+		}
+		if (query.getRouting() != null) {
+			indexRequestBuilder.setRouting(query.getRouting());
 		}
 
 		return indexRequestBuilder;
--- a/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
@@ -44,6 +44,7 @@ import org.springframework.data.elasticsearch.ElasticsearchException;
 import org.springframework.data.elasticsearch.annotations.ScriptedField;
 import org.springframework.data.elasticsearch.core.document.Document;
 import org.springframework.data.elasticsearch.core.document.SearchDocument;
+import org.springframework.data.elasticsearch.core.join.JoinField;
 import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentEntity;
 import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentProperty;
 import org.springframework.data.elasticsearch.core.mapping.ElasticsearchPersistentPropertyConverter;
@@ -709,6 +710,13 @@ public class MappingElasticsearchConverter
 
 			if (container.equals(type) && type.getType().equals(actualType)) {
 				return false;
+			}
+
+			if (container.getRawTypeInformation().equals(type)) {
+				Class<?> containerClass = container.getRawTypeInformation().getType();
+				if (containerClass.equals(JoinField.class) && type.getType().equals(actualType)) {
+					return false;
+				}
 			}
 		}
 
--- a/src/main/java/org/springframework/data/elasticsearch/core/join/JoinField.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/join/JoinField.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core.join;
+
+import org.springframework.lang.Nullable;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Subhobrata Dey
+ * @since 4.1
+ */
+public class JoinField<ID> {
+
+    private final String name;
+
+    @Nullable private ID parent;
+
+    public JoinField() {
+        this("default", null);
+    }
+
+    public JoinField(String name) {
+        this(name, null);
+    }
+
+    public JoinField(String name, @Nullable ID parent) {
+        this.name = name;
+        this.parent = parent;
+    }
+
+    public void setParent(@Nullable ID parent) {
+        this.parent = parent;
+    }
+
+    @Nullable
+    public ID getParent() {
+        return parent;
+    }
+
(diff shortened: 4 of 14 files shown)


sothawo:
thanks for the PR; I will have time to have a closer look at it the next days.
From a first glimpse:

sbcd90:
Hi @sothawo , Thanks for your comments on this PR. I have resolved the following issues:

- Dependencies removed from pom.xml
- Removed jackson objectmapper from mapping classes.

sothawo:
Sorry it took me this long, I first had to find the time to get into this topic. Thanks for your work, there is quite some stuff that's worth taking, but I think that in some parts we should go a different way. Let me explain what I mean by this:

## what we should aim for

In the examples I use the following relationships

* a _question_ can be parent of an _answer_ or a _comment_.
* an _answer_ can be the parent of a _vote_.

For the properties in an entity we should provide a `JoinField<ID>` class which might look like this:

```java
public class JoinField<ID> {
	private final String name;
	@Nullable private ID parent;

    // constructors, getter, setter etc.
}
```

This is a generic class, because the `parent` value needs to be of the same type as the ID property of the entity class.

In addition to that, we need some annotations to be able to define the parent-child relationships.
The target should be that the user can define the join-type on an entity like this:

```java
@Document(indexName="my-index")
public class JoinFieldDocument {
	@Id @Field(type = Keyword) private String id;

	@Field(type = Text) private String text;

	@JoinTypeRelations(relations = {
			@JoinTypeRelation(parent = "question", children = { "answer", "comment" }),
			@JoinTypeRelation(parent = "answer", children = { "vote" })
	})
	private JoinField<String> relation;
}
```

## building the index mappings

The `JoinField` class and the annotations is all the `MappingBuilder` would need to write the correct index mappings. As a reference you might check how `GeoPoint` properties are processed in `MappingBuilder.buildPropertyMapping(XContentBuilder, boolean, ElasticsearchPersistentProperty)`.

## writing and reading the entity when indexing or retrieving the document

If the ID type of the entity always would be a `String` then nothing would need to be changed. But as this could be any type, there are possible conversions involved. I need to check at which place in the  `MappingElasticsearchConverter` this would be best done.

## building queries for indexing, update and delete

Here we need - probably in the `RequestFactory` a check, if the `ElasticsearchPersistentEntity` has a joinfield property (and only here it is needed, not for mappings or converting). If so, then the `parent` property of this `JoinField` property must be used to add the _routing_ for the request. The routing must not be defined by an additional property,
because it must match the _parent_ entry of the `JoinField`

## summary

a lot of text for a review, I hope I could make clear where the basic changes need to be made. Please let me now, if this approach makes sense and what you think of it.

sbcd90:
Hi @sothawo , Extremely sorry for delay in my response. The proposal really looks good & I made the changes accordingly.
Kindly have a look.

src/main/java/org/springframework/data/elasticsearch/core/query/IndexQuery.java
```diff
@@ -107,4 +108,13 @@ public Long getPrimaryTerm() {
 	public void setPrimaryTerm(Long primaryTerm) {
 		this.primaryTerm = primaryTerm;
 	}
+
+	@Nullable
+	public String getRouting() {
+		return routing;
+	}
+
+	public void setRouting(String routing) {
```
sothawo, line 117: parameter should be annotated with `@Nullable`

src/main/java/org/springframework/data/elasticsearch/core/query/IndexQueryBuilder.java
```diff
@@ -66,6 +67,11 @@ public IndexQueryBuilder withSeqNoPrimaryTerm(SeqNoPrimaryTerm seqNoPrimaryTerm)
 		return this;
 	}
 
+	public IndexQueryBuilder withRouting(String routing) {
```
sothawo, line 70: parameter should be annotated with `@Nullable`

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
```diff
@@ -576,6 +576,10 @@ public IndexRequest indexRequest(IndexQuery query, IndexCoordinates index) {
 			indexRequest.setIfPrimaryTerm(query.getPrimaryTerm());
 		}
 
+		if (query.getRouting() != null) {
```
sothawo, line 579: the routing must be added as well to the method `indexRequestBuilder` for the transport client.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
```diff
@@ -3013,6 +3017,61 @@ void shouldSaveEntityWithIndexCoordinates() {
...
+
+		JoinField<String> myAJoinField2 = new JoinField<>("answer");
+		myAJoinField2.setParent(qId1);
+		sampleAnswerEntity2.setMyJoinField(myAJoinField2);
+
+		operations.save(Arrays.asList(sampleQuestionEntity1, sampleQuestionEntity2,
+				sampleAnswerEntity1, sampleAnswerEntity2), IndexCoordinates.of(INDEX_NAME_JOIN_SAMPLE_ENTITY));
+		indexOperations.refresh();
+
+		Thread.sleep(5000);
```
sothawo, line 3060: should not be necessary when doing a refresh

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -518,6 +520,33 @@ private String getEntityId(Object entity) {
 		return null;
 	}
 
+	@Nullable
+	private String getEntityRouting(Object entity) {
```
sothawo, line 524: the logic should go into the `ElasticsearchPersistentEntity`. We should add a method `boolean hasJoinFieldProperty()` and `ElasticsearchPersistenProperty getJoinFieldProperty()`. This would be initialized in `org.springframework.data.elasticsearch.core.mapping.SimpleElasticsearchPersistentEntity#addPersistentProperty`. The check if a property matches should check the type to be `JoinField`,  not the annotations.

Here in getEntityRouting we'd get that property (like it's done in  `getEntityVersion()`, then retrieve the `parent`value as Object and use `entityConverter.convertId()` to get the String for the routing value.

sbcd90:
hi @sothawo, i'll look into the test failures.

sbcd90:
hi @sothawo, i'll look into the test failures.

sbcd90:
hi @sothawo , i have made all the changes. The tests should pass now.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplateTests.java
```diff
@@ -119,4 +122,72 @@ void shouldUseAllOptionsFromUpdateQuery() {
 		assertThat(fetchSourceContext.excludes()).containsExactlyInAnyOrder("excl");
 	}
 
+	@Test
```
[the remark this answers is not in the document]
sbcd90: Unfortunately, the transport client doesnt support join datatype yet. Hence, tests run  only for the rest client.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
```diff
@@ -131,6 +133,12 @@ public void before() {
 		IndexOperations indexOpsSearchHitsEntity = operations.indexOps(SearchHitsEntity.class);
 		indexOpsSearchHitsEntity.create();
 		indexOpsSearchHitsEntity.putMapping(SearchHitsEntity.class);
+
```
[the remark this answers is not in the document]
sbcd90: I didnt understand the base class  part here. Unfortunately, the transport client doesnt support join datatype yet. Hence, indexes are required to run only for the rest client.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplateTests.java
```diff
@@ -119,4 +122,72 @@ void shouldUseAllOptionsFromUpdateQuery() {
...
+
+		assertThat(hitIds.containsAll(Arrays.asList(aId1, aId2))).isTrue();
+	}
+
+	@Data
+	@NoArgsConstructor
+	@AllArgsConstructor
+	@Builder
+	@Document(indexName = INDEX_NAME_JOIN_SAMPLE_ENTITY)
+	static class SampleJoinEntity {
```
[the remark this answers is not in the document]
sbcd90: Unfortunately, the transport client doesnt support join datatype yet. Hence, tests run only for the rest client.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchRestTemplateTests.java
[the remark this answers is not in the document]
sothawo: That would seem strange, the join type was introduced with Elasticsearch 6.0, when the old parent field was removed.  What errors do you get when running these tests with the transport client?

sothawo:
I just locally moved the test to the base class and ran it. As I wrote:

> Please change in the class org.springframework.data.elasticsearch.Utils line 58 to:
Arrays.asList(Netty4Plugin.class, ParentJoinPlugin.class)); otherwise the embedded Elasticsearch instance used for testing cannot work with join types.

The transport client test is failing, as the routing is not set, but that I already mentioned.

sbcd90:
Hi @sothawo , Thanks a lot for reviewing this.

Actually I'm sorry. My mistake. I should have taken care of the 2 points.

- implementation in the reactive code
- the routing must be considered for updates and deletes as well.

I have just 1 question. In the following line,
https://github.com/spring-projects/spring-data-elasticsearch/blob/2d0acbd61b85cb5b8404d1e465c7517f64274637/src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java#L478

`entity.getTypeInformation()` only gives back the `rawType` & not the `parameterizedType`.
Can this be handled to give `parameterizedType`?
Then, we dont need changes in `requiresTypeHint`.

sothawo:
I don't see why we need some changes in `requiresTypeHint()` at all. The`JoinField` itself doesn't need one and will be written and read without problems. As for the `parent` property which is generic: This is the same type as the id property of the entity (which we could check when building the persistent entity) and so it must b e either a type that can be converted by Spring Data Elasticseacrh (like String, UUID, Long, etc.) or the user must register a custom converter anyways for the id property and this will then be used for the `JoinType`as well.

sbcd90:
Hi @sothawo , Yes. The `JoinField` is read without any problems but writing has 1 problem.
Join doesnt need `type key` or `_class` property. 

The `ElasticsearchTypeMapper` method `create` defaults it to `_class`. I cannot set `typeKey` to null.

src/main/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverter.java
```diff
@@ -710,6 +711,13 @@ private boolean requiresTypeHint(TypeInformation<?> type, Class<?> actualType,
 			if (container.equals(type) && type.getType().equals(actualType)) {
 				return false;
 			}
+
+			if (container.getRawTypeInformation().equals(type)) {
```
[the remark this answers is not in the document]
sothawo: When I first tried my changes there were no type hints written, but testing this now they are, so this is needed. So that's ok here

sothawo:
Do you need any further support on this PR?

sbcd90:
Hi @sothawo , just working on this PR. I'll post here once I need support.

sbcd90:
Hi @sothawo , I worked on the following issues:

- the routing must be considered for updates and deletes as well.
- code reviews.

I would request you to kindly review the PR again. I'll only work on the remaining part if this PR looks ok.
What is remaining:

- implementation in the reactive code

src/main/java/org/springframework/data/elasticsearch/core/join/JoinField.java
```diff
@@ -0,0 +1,61 @@
...
+ * @author Subhobrata Dey
+ * @since 4.1
+ */
+public class JoinField<ID> {
+
+    private String name;
+
+    @Nullable private ID parent;
+
+    public JoinField() {
```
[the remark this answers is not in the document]
sbcd90: I get this error if I remove the no-args constructor.

```
org.springframework.data.mapping.model.MappingInstantiationException: Failed to instantiate org.springframework.data.elasticsearch.core.join.JoinField using constructor NO_CONSTRUCTOR with arguments 

	at org.springframework.data.mapping.model.ReflectionEntityInstantiator.createInstance(ReflectionEntityInstantiator.java:65)
	at org.springframework.data.mapping.model.ClassGeneratingEntityInstantiator.createInstance(ClassGeneratingEntityInstantiator.java:87)
	at org.springframework.data.elasticsearch.core.convert.MappingElasticsearchConverter.readEntity(MappingElasticsearchConverter.java:186)
	at org.springframework.data.elasticsearch.core.convert.MappingElasticsearchConverter.readMapValue(MappingElasticsearchConverter.java:357)
```

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
```diff
@@ -888,6 +892,10 @@ public IndexRequest indexRequest(IndexQuery query, IndexCoordinates index) {
...
```
sothawo, line 895: this needs to be added to `deleteByQueryRequestBuilder()` as well.

sbcd90: Hi @sothawo , i think `routing` is already present in `updateRequest`
https://github.com/spring-projects/spring-data-elasticsearch/blob/cee03ee3b95ebae6dad9bff49322131e339ecaf3/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java#L1361-#L1363
& for `updateRequestBuilder`
https://github.com/spring-projects/spring-data-elasticsearch/blob/cee03ee3b95ebae6dad9bff49322131e339ecaf3/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java#L1435-#L1437

sbcd90:
Hi @sothawo , i added `routing` to `deleteByQueryRequestBuilder()` as well.

sothawo:
Great. One last question: Can you please rebase your branch on master? We had a change today - that does not affect your code - but the dependencies in maven.

I would then merge this PR.

As for the reactive part: Let's create a new issue for that. I'd rather have this PR merged, it has been open quite long. I see no problem if the next milestone release (planned for August 12th) does not yet have support for the join-type in the reactive code, there's more than enough time to add this to RC1 or RC2;

sbcd90:
Hi @sothawo , rebased the PR.

sothawo:
thanks again for this PR and the time you took for this. I will later create a ticket for adding this to the reactive part.

PR #498, opened 2020-08-05
DATAES-894 - Adapt to changes in Reactor

bclozel:
See DATAES-893


mp911de:
Thank you for your contribution. That's merged and polished now.

PR #506, opened 2020-08-18
DATAES-908 - Fill version on an indexed entity.

rpuch:
Elasticsearch returns version of a document after it has been indexed, so we can (and probably should) fill it on the entity, like we do it for SeqNoPrimaryTerm (this behavior was added in DATAES-876).

Also, SeqNoPrimaryTerm was not filled in a couple of places on indexing.

This still has some issues.

1. 2 tests fail, namely ElasticsearchTemplateTests.shouldReturnObjectsForGivenIdsUsingMultiGet() and ElasticsearchTemplateTests.shouldReturnNullObjectForNotExistingIdUsingMultiGet(). Both seem to check for too much: they both require the version field to remain unchanged. According to their names, I would simply remove the version field from the test entity. On the other hand, if an intention of the tests is to make sure that the @Version property is actually providing the value for the version that should be used by ES to index the documents, this should be discussed and fixed, because spring-data-elasticsearch currently does not do this (and in my opinion, it shouldn't).
What would you advise here?
2. `IndexedObjectInformation` seems to have a problem: all its fields are nullable, and in one case they are really initialized with nulls, but all the getters return primitive values. Should that null-initializing place (`AbstractElasticsearchTemplate:499`) be rewritten to avoid such null initialization?

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
@@ -493,9 +493,10 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 		return Stream.of(bulkResponse.getItems()).map(bulkItemResponse -> {
 			DocWriteResponse response = bulkItemResponse.getResponse();
 			if (response != null) {
-				return IndexedObjectInformation.of(response.getId(), response.getSeqNo(), response.getPrimaryTerm());
+				return IndexedObjectInformation.of(response.getId(), response.getSeqNo(), response.getPrimaryTerm(),
+						response.getVersion());
 			} else {
-				return IndexedObjectInformation.of(bulkItemResponse.getId(), null, null);
+				return IndexedObjectInformation.of(bulkItemResponse.getId(), null, null, null);
 			}
 
 		}).collect(Collectors.toList());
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -662,6 +668,20 @@ protected void maybeCallbackAfterSaveWithQueries(List<?> queries, IndexCoordinat
 
 	// endregion
 
+	protected void updateIndexedObjectsWithQueries(List<?> queries,
+			List<IndexedObjectInformation> indexedObjectInformations) {
+		for (int i = 0; i < queries.size(); i++) {
+			Object query = queries.get(0);
```
sothawo, line 674: if you use `i` instead of `0` it works and the tests will be green. 😉

src/main/java/org/springframework/data/elasticsearch/core/IndexedObjectInformation.java
```diff
@@ -49,4 +54,8 @@ public long getSeqNo() {
 	public long getPrimaryTerm() {
 		return primaryTerm;
 	}
+
+	public long getVersion() {
```
sothawo, line 58: we should change this - and the other two getters to `@Nullable public Long` and return objects instead of primitives

sothawo (changes requested):
thanks for cleaning this up. your issues are addressed by my two comments in the files.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
rpuch: Oops, that's embarassing :) Night coding...

I've fixed this and also updated my tests to catch it as well.

src/main/java/org/springframework/data/elasticsearch/core/IndexedObjectInformation.java
rpuch: Done

PR #535, opened 2020-10-10
DATAES-934 - Add a Query taking method to ElasticsearchRepository.

sothawo:


src/main/java/org/springframework/data/elasticsearch/repository/ElasticsearchRepository.java
```diff
@@ -55,23 +56,32 @@
...
+	 *             annotated methods, or {@link org.springframework.data.elasticsearch.core.ElasticsearchOperations}.
 	 */
 	Page<T> search(Query searchQuery);
 
+	/**
+	 * execute the given query and return the result in a SearchPage.
+	 * 
+	 * @since 4.1
+	 */
+	SearchPage<T> searchQuery(Query query);
```
mp911de, line 83: We should get actually rid of these methods entirely as that's an anti-pattern in Spring Data. See also https://jira.spring.io/browse/DATAMONGO-2610.

PR #538, opened 2020-10-15
DATAES-953 - DateTimeException on converting Instant or Date to custo…

sothawo:
…m format.


zhouxin2333:
@sothawo I just woke up and I was happy to learn about this commit. 
Actually, my code now uses customizing SimpleElasticsearchMappingContext so that I can finally modify the parse method of ElasticsearchDateConverter. That can also work!

PR #554, opened 2020-11-13
DATAES-976 - Implement CrudRepository.delete(Iterable<ID> ids).

schauder:
Merge directly after DATACMNS-800

### DIFF
--- a/src/test/java/org/springframework/data/elasticsearch/repository/support/SimpleReactiveElasticsearchRepositoryTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/repository/support/SimpleReactiveElasticsearchRepositoryTests.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.data.elasticsearch.repository.support;
 
+import static java.util.Arrays.*;
 import static org.assertj.core.api.Assertions.*;
 import static org.springframework.data.elasticsearch.annotations.FieldType.*;
 import static org.springframework.data.elasticsearch.core.query.Query.*;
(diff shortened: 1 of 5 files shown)


src/test/java/org/springframework/data/elasticsearch/repository/support/SimpleReactiveElasticsearchRepositoryTests.java
```diff
@@ -107,7 +109,7 @@ public void saveShouldSaveSingleEntity() {
 	public void saveShouldComputeMultipleEntities() {
 
 		repository
-				.saveAll(Arrays.asList(SampleEntity.builder().build(), SampleEntity.builder().build(),
+				.saveAll(asList(SampleEntity.builder().build(), SampleEntity.builder().build(),
```
mp911de, line 112: Nit: We generally don't import `Arrays` statically to avoid clashes.

mp911de (approved):
Looks good. One thing that I noticed is that `deleteAll(…)` doesn't seem to make use of entity callbacks or lifecycle events @sothawo.

src/test/java/org/springframework/data/elasticsearch/repository/support/SimpleReactiveElasticsearchRepositoryTests.java
schauder: Hmm, I only knew the rule that static imports are ok in tests. Will undo

schauder: I reverted those changes.

mp911de:

PR #562, opened 2020-11-30
#1218 - Add routing parameter to ElasticsearchOperations.

sothawo:
The `MappingContext` now has a `RoutingAccessorFactory`.

The `@Document` annotation has a `routing` parameter (I call that the routing specification) that will be passed together with the bean to the `RoutingAccessorFactory` which then can produce a `RoutingAccessor` for the bean.

The default implementation first checks if the routing specification is the name of an `PersistentProperty` for the bean. If so, the value of that property will be used for routing. If not, the routing specification will be evaluated as a SpEL expression with the bean as root object, allowing to define some bean that will resolve to a routing value.

If this is not enough, a user can set a custom `RoutingAccessorFactory` in the `ConfigurationSupport`.

Where a routing is needed, but no bean/entity available (get/delete by id), I introduced additional methods taking a _routing_ parameter.

### DIFF
--- a/src/main/asciidoc/reference/elasticsearch-join-types.adoc
+++ b/src/main/asciidoc/reference/elasticsearch-join-types.adoc
@@ -1,0 +1,229 @@
+[[elasticsearch.jointype]]
+= Join-Type implementation
+
+Spring Data Elasticsearch supports the https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html[Join data type] for creating the corresponding index mappings and for storing the relevant information.
+
+== Setting up the data
+
+For an entity to be used in a parent child join relationship, it must have a property of type `JoinField` which must be annotated.
+Let's assume a `Statement` entity where a statement may be a _question_, an _answer_, a _comment_ or a _vote_ (a _Builder_ is also shown in this example, it's not necessary, but later used in the sample code):
+
+====
+[source,java]
+----
+@Document(indexName = "statements")
+@Routing("routing")                                                                       <.>
+public class Statement {
+    @Id
+    private String id;
+
+    @Field(type = FieldType.Text)
+    private String text;
+
+    @Field(type = FieldType.Keyword)
+    private String routing;
+
+    @JoinTypeRelations(
+        relations =
+            {
+                @JoinTypeRelation(parent = "question", children = {"answer", "comment"}), <.>
+                @JoinTypeRelation(parent = "answer", children = "vote")                   <.>
+            }
+    )
+    private JoinField<String> relation;                                                   <.>
+
+    private Statement() {
+    }
+
+    public static StatementBuilder builder() {
+        return new StatementBuilder();
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public String getRouting() {
+        return routing;
+    }
+
+    public void setRouting(Routing routing) {
+        this.routing = routing;
+    }
+
+    public String getText() {
+        return text;
+    }
+
+    public void setText(String text) {
+        this.text = text;
+    }
+
+    public JoinField<String> getRelation() {
+        return relation;
+    }
+
+    public void setRelation(JoinField<String> relation) {
+        this.relation = relation;
+    }
+
+    public static final class StatementBuilder {
+        private String id;
+        private String text;
+        private String routing;
+        private JoinField<String> relation;
+
+        private StatementBuilder() {
+        }
+
+        public StatementBuilder withId(String id) {
+            this.id = id;
+            return this;
+        }
+
+        public StatementBuilder withRouting(String routing) {
+            this.routing = routing;
+            return this;
+        }
+
+        public StatementBuilder withText(String text) {
+            this.text = text;
+            return this;
+        }
+
+        public StatementBuilder withRelation(JoinField<String> relation) {
+            this.relation = relation;
+            return this;
+        }
+
+        public Statement build() {
+            Statement statement = new Statement();
+            statement.setId(id);
+            statement.setRouting(routing);
+            statement.setText(text);
+            statement.setRelation(relation);
+            return statement;
+        }
+    }
+}
+----
+<.> for routing related info see <<elasticsearch.routing>>
+<.> a question can have answers and comments
+<.> an answer can have votes
+<.> the `JoinField` property is used to combine the name (_question_, _answer_, _comment_ or _vote_) of the relation with the parent id.
+The generic type must be the same as the `@Id` annotated property.
+====
+
+Spring Data Elasticsearch will build the following mapping for this class:
+
+====
+[source,json]
+----
+{
+  "statements": {
+    "mappings": {
+      "properties": {
+        "_class": {
+          "type": "text",
+          "fields": {
+            "keyword": {
+              "type": "keyword",
+              "ignore_above": 256
+            }
+          }
+        },
+        "routing": {
+          "type": "keyword"
+        },
+        "relation": {
+          "type": "join",
+          "eager_global_ordinals": true,
+          "relations": {
+            "question": [
+              "answer",
+              "comment"
+            ],
+            "answer": "vote"
+          }
+        },
+        "text": {
+          "type": "text"
+        }
+      }
+    }
+  }
+}
+----
+====
+
+==  Storing data
+
+Given a repository for this class the following code inserts a question, two answers, a comment and a vote:
+
+====
+[source,java]
+----
+void init() {
+    repository.deleteAll();
+
+    Statement savedWeather = repository.save(
+        Statement.builder()
+            .withText("How is the weather?")
+            .withRelation(new JoinField<>("question"))                          <1>
+            .build());
+
+    Statement sunnyAnswer = repository.save(
+        Statement.builder()
+            .withText("sunny")
+            .withRelation(new JoinField<>("answer", savedWeather.getId()))      <2>
+            .build());
+
+    repository.save(
+        Statement.builder()
+            .withText("rainy")
+            .withRelation(new JoinField<>("answer", savedWeather.getId()))      <3>
+            .build());
+
+    repository.save(
+        Statement.builder()
+            .withText("I don't like the rain")
+            .withRelation(new JoinField<>("comment", savedWeather.getId()))     <4>
+            .build());
+
+    repository.save(
+        Statement.builder()
+            .withText("+1 for the sun")
+            ,withRouting(savedWeather.getId())
+            .withRelation(new JoinField<>("vote", sunnyAnswer.getId()))         <5>
+            .build());
+}
+----
+<1> create a question statement
+<2> the first answer to the question
+<3> the second answer
+<4> a comment to the question
+<5> a vote for the first answer, this needs to have the routing set to the weather document, see <<elasticsearch.routing>>.
+====
+
+==  Retrieving data
+
+Currently native search queries must be used to query the data, so there is no support from standard repository methods. <<repositories.custom-implementations>> can be used instead.
+
+The following code shows as an example how to retrieve all entries that have a _vote_ (which must be _answers_, because only answers can have a vote) using an `ElasticsearchOperations` instance:
+
+====
+[source,java]
+----
+SearchHits<Statement> hasVotes() {
+    NativeSearchQuery query = new NativeSearchQueryBuilder()
+        .withQuery(hasChildQuery("vote", matchAllQuery(), ScoreMode.None))
+        .build();
+
+    return operations.search(query, Statement.class);
+}
+----
+====
--- a/src/main/asciidoc/reference/elasticsearch-routing.adoc
+++ b/src/main/asciidoc/reference/elasticsearch-routing.adoc
@@ -1,0 +1,106 @@
+
+[[elasticsearch.routing]]
+= Routing values
+
+When Elasticsearch stores a document in an index that has multiple shards, it determines the shard to you use based on the _id_ of the document.
+Sometimes it is necessary to predefine that multiple documents should be indexed on the same shard (join-types, faster search for related data).
+For this Elasticsearch offers the possibility to define a routing, which is the value that should be used to calculate the shard from instead of the _id_.
+
+Spring Data Elasticsearch supports routing definitions on storing and retrieving data in the following ways:
+
+== Routing on join-types
+
+When using join-types (see <<elasticsearch.jointype>>), Spring Data Elasticsearch will automatically use the `parent` property of the entity's `JoinField` property as the value for the routing.
+
+This is correct for all the use-cases where the parent-child relationship has just one level.
+If it is deeper, like a child-parent-grandparent relationship - like in the above example from _vote_ -> _answer_ -> _question_ - then the routing needs to explicitly specified by using the techniques described in the next section (the _vote_ needs the _question.id_ as routing value).
+
+== Custom routing values
+
+To define a custom routing for an entity, Spring Data Elasticsearch provides a `@Routing` annotation (reusing the `Statement` class from above):
+
+====
+[source,java]
+----
+@Document(indexName = "statements")
+@Routing("routing")                  <.>
+public class Statement {
+    @Id
+    private String id;
+
+    @Field(type = FieldType.Text)
+    private String text;
+
+    @JoinTypeRelations(
+        relations =
+            {
+                @JoinTypeRelation(parent = "question", children = {"answer", "comment"}),
+                @JoinTypeRelation(parent = "answer", children = "vote")
+            }
+    )
+    private JoinField<String> relation;
+
+    @Nullable
+    @Field(type = FieldType.Keyword)
+    private String routing;          <.>
+
+    // getter/setter...
+}
+----
+<.> This defines _"routing"_ as routing specification
+<.> a property with the name _routing_
+====
+
+If the `routing` specification of the annotation is a plain string and not a SpEL expression, it is interpreted as the name of a property of the entity, in the example it's the _routing_ property.
+The value of this property will then be used as the routing value for all requests that use the entity.
+
+It is also possible to us a SpEL expression in the `@Document` annotation like this:
+
+====
+[source,java]
+----
+@Document(indexName = "statements")
+@Routing("@myBean.getRouting(#entity)")
+public class Statement{
+    // all the needed stuff
+}
+----
+====
+
+In this case the user needs to provide a bean with the name _myBean_ that has a method `String getRouting(Object)`. To reference the entity _"#entity"_ must be used in the SpEL expression, and the return value must be `null` or the routing value as a String.
+
+If plain property's names and SpEL expressions are not enough to customize the routing definitions, it is possible to define provide an implementation of the `RoutingResolver` interface. This can then be set on the `ElasticOperations` instance:
+
+====
+[source,java]
+----
+RoutingResolver resolver = ...;
+
+ElasticsearchOperations customOperations= operations.withRouting(resolver);
+
+----
+====
+
+The `withRouting()` functions return a copy of the original `ElasticsearchOperations` instance with the customized routing set.
+
+
+When a routing has been defined on an entity when it is stored in Elasticsearch, the same value must be provided when doing a _get_ or _delete_ operation. For methods that do not use an entity - like `get(ID)` or `delete(ID)` - the `ElasticsearchOperations.withRouting(RoutingResolver)` method can be used like this:
+
+====
+[source,java]
+----
+String id = "someId";
+String routing = "theRoutingValue";
+
+// get an entity
+Statement s = operations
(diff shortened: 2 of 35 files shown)


sothawo:
@mp911de after our last call I reworked this.

The configuration of the routing is not an parameter in `@Document` anymore, but has been moved into a new `@Routing` annotation.

The value of this annotation is _not_ a fixed routing value. This would not make sense. `@Document` defines the index for an entity class. When this index has multiple shards, using a fixed routing value for all entities would store all entities in the same shard, having the same effect as only having one shard in the index.

Resolving the routing for an entity is done by a `RoutingResolver`, the default implementation does this:

* if the `@Routing` value is the name of a property, the value of this property is used as routing value
* otherwise it is evaluated a SpEL expression against the root context where the entity can be referenced with "#entity"

`ElasticsearchOperations` has no more overloads for `get` or `delete`,  but has two methods named `withRouting()`, one taking a String for a fixed routing value, the other a `RoutingResolver`.

sothawo:
@mp911de if you find the time to have a look…

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -868,4 +909,38 @@ public ReadSearchScrollDocumentResponseCallback(Class<T> type, IndexCoordinates
...
+	public void setRoutingResolver(RoutingResolver routingResolver) {
+
+		Assert.notNull(routingResolver, "routingResolver must not be null");
+
+		this.routingResolver = routingResolver;
+	}
+
+	@Override
+	public RoutingResolver getRoutingResolver() {
+		return Objects.requireNonNull(routingResolver);
```
mp911de, line 924: `Objects.requireNonNull` throws a `NPE`. How can it happen that `RoutingResolver` is null?

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -868,4 +909,38 @@ public ReadSearchScrollDocumentResponseCallback(Class<T> type, IndexCoordinates
...
+	}
+
+	@Override
+	public ElasticsearchOperations withRouting(RoutingResolver routingResolver) {
```
mp911de, line 928: Instead of additional methods, how about `RoutingResolver.just(String)`?

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
```diff
@@ -416,4 +417,37 @@ default String stringIdRepresentation(@Nullable Object id) {
 		return Objects.toString(id, null);
 	}
 	// endregion
+
+	//region routing
+	/**
+	 * Sets the current {@link RoutingResolver}.
+	 * @param routingResolver must not be {@literal null}
+	 */
+	void setRoutingResolver(RoutingResolver routingResolver);
```
mp911de, line 426: I'd suggest to not expose these methods on ther operations interface. They should not be used by the application in the first place.

src/main/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchOperations.java
```diff
@@ -88,6 +89,38 @@
 	 */
 	ReactiveIndexOperations indexOps(Class<?> clazz);
 
+	//region routing
+	/**
+	 * Sets the current {@link RoutingResolver}.
+	 * @param routingResolver must not be {@literal null}
+	 */
+	void setRoutingResolver(RoutingResolver routingResolver);
```
mp911de, line 97: See comments about these methods on `ElasticSearchOperations`.

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
```diff
@@ -171,9 +171,35 @@ default ElasticsearchPersistentProperty getRequiredSeqNoPrimaryTermProperty() {
...
+	/**
+	 * Resolves the routing for a bean in a SpEL context if {@link #hasRouting()} is true. The value of
+	 * {@link #getRouting()} is taken as a SpEL expression will be resolved to a String. The #bean is set as the context
+	 * root and can be referenced with '#root' or '#this'.
+	 * 
+	 * @param bean the bean to resolve the routing for
+	 * @return routing value, may be {@literal null}
+	 */
+	@Nullable
+	String resolveRouting(Object bean);
```
mp911de, line 204: The fact that routing can be SpEL should not be exposed. Evaluation should happen inside of the `PersistentEntity`.

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
```diff
@@ -171,9 +171,35 @@ default ElasticsearchPersistentProperty getRequiredSeqNoPrimaryTermProperty() {
...
 	 * @since 4.1
 	 */
 	Document getDefaultSettings();
+
+	/**
+	 * @return the routing specification for the entity
+	 * @since 4.2
+	 */
+	@Nullable
+	String getRouting();
```
mp911de, line 185: Should return the evaluated expression (when using SpEL) or the routing string (when using literals).

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	/**
+	 * First tries to resolve the #routing as a property of the bean. If this property exists, it's value is read. If this
+	 * value is not null, to String() is called on the value. If the configured routing is not a property, then a SpEL
+	 * resolution is attempted { @see {@link ElasticsearchPersistentEntity#resolveRouting(Object)}}
+	 */
+	@Override
+	@Nullable
+	public String getRouting(@Nullable Object bean, @Nullable String routing) {
```
mp911de, line 49: This method is either called without or with a bean for two well-defined cases. 

I'd suggest to have two distinct methods `getRouting(Class)` and `getRouting(Class, Object)`. Handing in the routing argument leaks the fact that routing can be defined on the type already. Maybe it would even make sense to have `getRouting(IndexCoordinates)` as all these calls to `routingResolver.getRouting(null, null)` are dealing with `null` in the first place.

src/main/java/org/springframework/data/elasticsearch/core/routing/package-info.java
```diff
@@ -0,0 +1,3 @@
+@org.springframework.lang.NonNullApi
```
mp911de, line 1: Nit: Package Javadoc

src/main/java/org/springframework/data/elasticsearch/core/routing/RoutingResolver.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core.routing;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * @author Peter-Josef Meisch
```
mp911de, line 21: Nit: `@since`

mp911de (changes requested):
This looks much cleaner now, good work. I left a few comments, mostly about the design of the routing resolver. Resolution means in that context to obtain the value from the context. I think splitting out more methods to make the routing more null-safe makes a lot of sense. Also, take a look at the design of `TypeAliasAccessor` and `Alias` to represent the absence of a defined routing key.

src/main/java/org/springframework/data/elasticsearch/core/routing/RoutingResolver.java
sothawo: done.

src/main/java/org/springframework/data/elasticsearch/core/routing/package-info.java
sothawo: done

src/main/asciidoc/reference/elasticsearch-routing.adoc
```diff
@@ -0,0 +1,102 @@
...
+If the `routing` specification of the annotation is a plain string and not a SpEL expression, it is interpreted as the name of a property of the entity, in the example it's the _routing_ property.
+The value of this property will then be used as the routing value for all requests that use the entity.
+
+It is also possible to us a SpEL expression in the `@Document` annotation like this:
+
+====
+[source,java]
+----
+@Document(indexName = "statements")
+@Routing("@spelRouting.getRouting(#entity)")
```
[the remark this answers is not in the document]
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
sothawo: it cannot happen from the program logic. The property is set during instance initialization and afterwards it cannot be set to null. But as it is not initialized in a constructor, the property is `@Nullable`.

The whole initialization code will get cleaner (and this class removed) once we have Elasticsearch 8 and the transport client removed. Then we can make these fields not nullable and initialize them in the constructor.

For now I removed the `Objects.requireNonNull()` and added a comment to keep the compiler happy instead of using rNN

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/ElasticsearchOperations.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchOperations.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
sothawo: No. this is the value of the `@Routing` annotation that is defined on the entity class. Do you mean this should return a SpEL expression?

sothawo: I removed this method, the persistent entity now only has the `resolveRouting()` method, that now contains the complete logic.

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
sothawo: I now have `getRouting()` for the fixed-value case and `getRouting(object bean)`. So it's no longer possible dot use a lambda for `routingResolver` in `operations.withRouting(routingResolver)`, but with `RoutingResolver.just(fixedValue)` this is ok.

sothawo:
@mp911de 👋  ready for another round…

src/main/asciidoc/reference/elasticsearch-join-types.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+== Setting up the data
+
+For an entity to be used in a parent child join relationship, it must have a property of type `JoinField` which must be annotated.
+Let's assume a `Statement` entity where a statement may be a _question_, an _answer_, a _comment_ or a _vote_ (a _Builder_ is also shown in this example, it's not necessary, but later used in the sample code):
+
+====
+[source,java]
+----
+@Document(indexName = "statements", routing="routing")                                    <.>
```
mp911de, line 14: Nit: Shouldn't that be `@Routing(…)`?

src/main/java/org/springframework/data/elasticsearch/core/routing/RoutingResolver.java
```diff
@@ -0,0 +1,62 @@
...
+	@Nullable
+	String getRouting(Object bean);
+
+	/**
+	 * Returns a {@link RoutingResolver that always retuns a fixed value}
+	 * 
+	 * @param value the value t return
+	 * @return the fixed-value {@link RoutingResolver}
+	 */
+	static RoutingResolver just(String value) {
```
mp911de, line 49: Nit: Add an assertion.

src/main/java/org/springframework/data/elasticsearch/core/routing/RoutingResolver.java
```diff
@@ -0,0 +1,62 @@
...
+	 * @param bean the bean to get the routing for
+	 * @return the routing value
+	 */
+	@Nullable
+	String getRouting(Object bean);
+
+	/**
+	 * Returns a {@link RoutingResolver that always retuns a fixed value}
+	 * 
+	 * @param value the value t return
```
mp911de, line 46: Nit: Typo

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
```diff
@@ -0,0 +1,64 @@
...
+	@Override
+	public String getRouting() {
+		return null;
+	}
+
+	@Override
+	@Nullable
+	public String getRouting(@Nullable Object bean) {
+
+		if (bean == null) {
```
mp911de, line 52: That should never happen that `bean` is `null`.

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Default implementation of the {@link RoutingResolver} interface. Retuns {@literal null} for the non-bean method and
+ * delegates to the corresponding persistent entity for the bean-method.
+ * 
+ * @author Peter-Josef Meisch
+ * @since 4.2
+ */
+public class DefaultRoutingResolver implements RoutingResolver {
+
+	private final SimpleElasticsearchMappingContext mappingContext;
```
mp911de, line 33: Let's use the `MappingContext<? extends ElasticsearchPersistentEntity, ? extends ElasticsearchPersistentProperty>` interface.

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
```diff
@@ -0,0 +1,64 @@
...
+
+	public DefaultRoutingResolver(MappingContext mappingContext) {
```
mp911de, line 35: Parameter should make use of generics.

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
```diff
@@ -0,0 +1,64 @@
...
+
+		Assert.isInstanceOf(SimpleElasticsearchMappingContext.class, mappingContext,
```
mp911de, line 37: Finally, no need for a concrete type assertion since the required methods are available on the interfaces.

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
```diff
@@ -171,9 +171,18 @@ default ElasticsearchPersistentProperty getRequiredSeqNoPrimaryTermProperty() {
...
 	Document getDefaultSettings();
+
+	/**
+	 * Resolves the routing for a bean.
+	 * 
+	 * @param bean the bean to resolve the routing for
+	 * @return routing value, may be {@literal null}
+	 */
+	@Nullable
+	String resolveRouting(Object bean);
```
mp911de, line 187: Actually, `bean` should be of type `T`. I think there's a similar constrain when obtaining `PersistentPropertyAccessor`. Interestingly, `isNew(…)` accepts `Object`.

Given the way how this method is used, it's fine to keep `Object` here and add a type assertion inside the implementation.

mp911de (commented):
I added a few minor things to consider. I'm still not decided whether SpEL resolution inside the `PersistentEntity` is the best approach vs. external annotation lookup. Let's stick with the current design and see how it goes.

src/main/asciidoc/reference/elasticsearch-join-types.adoc
sothawo: sure

src/main/java/org/springframework/data/elasticsearch/core/routing/RoutingResolver.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/routing/RoutingResolver.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
sothawo: changing this to `MappingContext<? extends ElasticsearchPersistentEntity<?>, ElasticsearchPersistentProperty>` is no different than using `SimpleElasticsearchMappingContext`. So I'd rather change the constructor parameter to `SimpleElasticsearchMappingContext` and remove the assert.

If we use `MappingContext<?,?>` we need a cast/check in `getRouting` for the class of the persistent entity. As `resolveRouting(bean)` is defined on `ElasticsearchPersistentEntity`. Using a `SimpleElasticsearchMappingContext` the return type of `getPersistentEntity()` is `SimpleElasticsearchPersistentEntity` and that's ok.

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
sothawo: assert to what type?

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
sothawo: done

src/main/java/org/springframework/data/elasticsearch/core/routing/DefaultRoutingResolver.java
sothawo: done

eaudet:
Waiting for this new feature. If I can help (testing) would be happy to. Thanks!

sothawo:
@mp911de care to have another look?

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
mp911de: That `bean` is instance of `T` (the entity type).

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -385,6 +400,36 @@ private EvaluationContext getIndexNameEvaluationContext() {
...
+		try {
+			Expression expression = routingExpressions.computeIfAbsent(routing, PARSER::parseExpression);
+			ExpressionDependencies expressionDependencies = ExpressionDependencies.discover(expression);
+
+			EvaluationContext context = getEvaluationContext(null, expressionDependencies);
+			context.setVariable("entity", bean);
+
+			return expression.getValue(context, String.class);
+		} catch (EvaluationException e) {
+			LOGGER.warn("could not resolve", e);
```
mp911de, line 427: How about throwing an exception instead of silently falling back to `null`?

mp911de (approved):
Great work, this looks pretty decent.

PR #1688, opened 2021-02-11
Add Rescore functionality

### LINKED ISSUE #1686: Add rescore functionality to NativeSearchQuery
Hi, wanted to ask if there is any reason why the [rescore](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-rescore.html) functionality is not implemented for NativeSearchQuery? Would you like a PR?

peermuellerxw:
Closes #1686

- [ x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ x] There is a ticket in the bug tracker for the project in our [issue tracker](https://github.com/spring-projects/spring-data-elasticsearch/issues).
- [ x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ x] You submit test cases (unit or integration tests) that back your changes.
- [ x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 import org.elasticsearch.action.DocWriteRequest;
@@ -37,7 +38,6 @@ import org.elasticsearch.action.admin.indices.exists.indices.IndicesExistsReques
 import org.elasticsearch.action.admin.indices.mapping.put.PutMappingRequestBuilder;
 import org.elasticsearch.action.admin.indices.refresh.RefreshRequest;
 import org.elasticsearch.action.admin.indices.settings.get.GetSettingsRequest;
-import org.elasticsearch.action.admin.indices.settings.get.GetSettingsResponse;
 import org.elasticsearch.action.admin.indices.template.delete.DeleteIndexTemplateRequest;
 import org.elasticsearch.action.bulk.BulkRequest;
 import org.elasticsearch.action.bulk.BulkRequestBuilder;
@@ -65,7 +65,6 @@ import org.elasticsearch.client.indices.IndexTemplatesExistRequest;
 import org.elasticsearch.client.indices.PutIndexTemplateRequest;
 import org.elasticsearch.client.indices.PutMappingRequest;
 import org.elasticsearch.common.geo.GeoDistance;
-import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.unit.DistanceUnit;
 import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.index.VersionType;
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@peermuellerxw Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=1688)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=1688) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@peermuellerxw Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-elasticsearch&pullRequestId=1688)!

src/main/java/org/springframework/data/elasticsearch/core/query/RescorerQuery.java
```diff
@@ -0,0 +1,40 @@
...
+ * classes into the query API.
+ *
+ * @author Peer Mueller
+ * @since 4.2
+ */
+public class RescorerQuery {
+
+	private final QueryRescorerBuilder queryRescorerBuilder;
+
+	public RescorerQuery(QueryRescorerBuilder queryRescorerBuilder) {
```
sothawo, line 31: this is leaking the Elasticsearch class into the API

sothawo (changes requested):
Thanks for this PR, great work, but I think we should take this further. It is focused on the `NativeSearchQuery`, but I think we should consider the other `Query` implementations here as well.

Something like having a 
```java
public class RescorerQuery {
  Query query; // <-  our Query interface
  ScoreMode scoreMode;
  int windowSize;
}
```
not using Elasticsearch classes here, but also having our own enum for the ScoreMode. This would allow to not only use `NativeSearchQuery` but also for example a `StringQuery`. In the `RequestMapper` this would then be mapped to the corresponding Elasticsearch classes – there are existing methods to convert a `Query` to a `QueryBuilder`.

What do you think about this change?

peermuellerxw:
Thanks for looking into my PR and for the feedback. Your suggestions do make sense and I am looking to implement them in the following days :)

peermuellerxw:
I updated the PR and changed the `RescorerQuery` to a "proper" domain object. Let me know if this is what you had in mind and also if you have any other comments/suggestions. @sothawo

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
```diff
@@ -1260,6 +1269,30 @@ private void prepareSort(Query query, SearchRequestBuilder searchRequestBuilder,
 			}
 		}
 	}
+
+	private QueryRescorerBuilder getQueryRescorerBuilder(RescorerQuery rescorerQuery) {
+
+		QueryRescorerBuilder builder = new QueryRescorerBuilder(Objects.requireNonNull(getQuery(rescorerQuery.getQuery())));
```
sothawo, line 1275: Instead of a `NullPointerException` we should throw an `IllegalArgumentException` here

sothawo (approved):
Looks fine, there are just some minor points with null checks. No need for you to fix these, I will do this in a follow-up commit after merging this PR.

Issue #1646, opened 2021-01-15
Add function to list all indexes


jbauerrfid:
Please add a function to list all indexes on the Elasticsearch host to `ReactiveElasticsearchClient.indices()`.

This should do the same as running `GET /_all` on the ES host, and return a Flux of index descriptions.

jbauerrfid:
as a work around I build a query on top of .execute to fetch the index list:

```
esClient.execute(cb -> cb.get()
                .uri("/" + indexName)
                .accept(MediaType.APPLICATION_JSON)
                .exchange())
                .flatMap(cr -> {
                    if (clientResponse.statusCode() == HttpStatus.OK) {
                        return clientResponse.bodyToMono(ObjectNode.class);
                    }
                    return Mono.empty();
                })
                .flatMapIterable(o  -> {
                   // convert results and collect them into a Flux...
                }
```

it does its job, but it feels a bit weird to do a lot of special logic for a pretty basic task

gpopides:
I tried to look at the code in order to grasp whats going on.

spring-data-elasticsearch uses response classes from elasticsearch library, so there is not a response which can represent the response of ```"/_all"``` endpoint.

I tried to map it my self but  i have troubles reading the response body from 
```java 
RawActionResponse
```

Any hint would be welcome, and then i could open a PR

sothawo:
__all_  is just a special case of the [get index API](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/indices-get-index.html). 

We should have a new function in `IndexOperations` like `getInformation(IndexCoordinates)` (and a corresponding function in the reactive part). The parameter could then contain a single index name, multiple names or a name like __all_ or _*_. The return value could be a list of some new `IndexInformation` class that contains the information for one specific index.

To get the result for this call it is necessary to use a `GetIndexRequest` with the `RestHighlevelClient` with the corresponding index names. The `RestHighlevelClient`then returns a `org.elasticsearch.client.indices.GetIndexResponse` object which contains (taken from the code):

```
    private Map<String, MappingMetadata> mappings;
    private Map<String, List<AliasMetadata>> aliases;
    private Map<String, Settings> settings;
    private Map<String, Settings> defaultSettings;
    private Map<String, String> dataStreams;
    private String[] indices;

```
and the needed getter functions to retrieve the information for all the indices. This information should be used to build the  `IndexInformation` instances.

[...]

PR #1693, opened 2021-02-14
Add function to list all indexes

gpopides:
- [ x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ ] There is a ticket in the bug tracker for the project in our [issue tracker](https://github.com/spring-projects/spring-data-elasticsearch/issues).
- [ x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ x] You submit test cases (unit or integration tests) that back your changes.
- [x ] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

This PR is for #1646.

I don't know if there is a chance where the new ```IndexInformation``` class would be delivered directly to a client, but if there is a such chance, this will fail.
For some reason, Jackson can not encode the ```settings``` attrribute (Settings class) with the following message: 

> Failed to get setting group for [] setting prefix and setting [refresh_interval] because of a missing '.'

for the following json:
```json
{
  "INDEX_NAME": {
    "aliases": {},
    "mappings": {},
    "settings": {
      "index": {
        "refresh_interval": "1s",
        "number_of_shards": "1",
        "provided_name": "X",
        "creation_date": "1613299161583",
        "store": {
          "type": "fs"
        },
        "number_of_replicas": "1",
        "uuid": "swcy7LHKSEC8kJpWBkGIJw",
        "version": {
          "created": "6080099"
        }
      }
    }
  }
}

```

Also, ```IndexInformation``` has some kind of duplicated code, but that's because of the TransportClient GetIndexRequest.
When it get's deprecated, i can clean it.

I would be happy to hear comments since this is my first PR.


src/main/java/org/springframework/data/elasticsearch/core/mapping/IndexInformation.java
```diff
@@ -0,0 +1,120 @@
...
+
+	private IndexInformation(String name, Settings settings, MappingMetadata mappingMetadata, List<AliasMetadata> aliases) {
+		this.name = name;
+		this.settings = settings;
+		this.mappings = mappingMetadata;
+		this.aliases = aliases;
+	}
+
+	@Override
+	public String toString() {
```
sothawo, line 105: If we need a `toString()` it should contain the content of the properties as well.

src/main/java/org/springframework/data/elasticsearch/core/mapping/IndexInformation.java
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @author George Popides
+ * @since 4.2
+ */
+public class IndexInformation {
+	private final String name;
+	private final Settings settings;
+	private final MappingMetadata mappings;
+	private final List<AliasMetadata> aliases;
+
```
sothawo, line 39: settings, mapping and alias information should probably be annotated as `@Nullable` as they need not be set ion an index.

In Spring Data Elasticsearch every parameter, property or return type that is not annotated with `org.springframework.lang.Nullable` is regarded as being non-null.

src/main/java/org/springframework/data/elasticsearch/core/IndexOperations.java
```diff
@@ -317,5 +318,7 @@ default boolean deleteTemplate(String templateName) {
 	 */
 	IndexCoordinates getIndexCoordinates();
 
+	List<IndexInformation> getInformation();
```
sothawo, line 321: please add Javadoc with a `@since 4.2` tag

src/main/java/org/springframework/data/elasticsearch/core/ReactiveIndexOperations.java
```diff
@@ -284,5 +286,6 @@
 	 */
 	IndexCoordinates getIndexCoordinates();
 
+	Mono<List<IndexInformation>> getInformation();
```
sothawo, line 289: we should rather return a `Flux<IndexInformation>` than a Mono of a list here.
Please add Javadoc and `@since 4.2`

src/main/java/org/springframework/data/elasticsearch/core/DefaultTransportIndexOperations.java
```diff
@@ -296,4 +300,16 @@ public boolean deleteTemplate(DeleteTemplateRequest deleteTemplateRequest) {
...
+
+	@Override
+	public List<IndexInformation> getInformation() {
+		GetIndexRequest getIndexRequest = new GetIndexRequest();
+		IndexCoordinates index = getIndexCoordinates();
+
+		getIndexRequest.indices(index.getIndexNames());
+
+		GetIndexResponse response = client.admin().indices().getIndex(getIndexRequest).actionGet();
+
```
sothawo, line 312: here the mapping should be done from Elasticsearch types to our data

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
```diff
@@ -256,5 +259,16 @@ public boolean deleteTemplate(DeleteTemplateRequest deleteTemplateRequest) {
 				client -> client.indices().deleteTemplate(deleteIndexTemplateRequest, RequestOptions.DEFAULT).isAcknowledged());
 	}
 
+	@Override
+	public List<IndexInformation> getInformation() {
+		GetIndexRequest request = requestFactory.getIndexRequest(getIndexCoordinates());
+
+		return restTemplate.execute(
+				client -> {
+					GetIndexResponse getIndexResponse = client.indices().get(request, RequestOptions.DEFAULT);
```
sothawo, line 268: here the mapping should be done from Elasticsearch types to our data

src/main/java/org/springframework/data/elasticsearch/core/DefaultReactiveIndexOperations.java
```diff
@@ -304,6 +307,14 @@ public IndexCoordinates getIndexCoordinates() {
 		return (boundClass != null) ? getIndexCoordinatesFor(boundClass) : boundIndex;
 	}
 
+	@Override
+	public Mono<List<IndexInformation>> getInformation() {
+		org.elasticsearch.client.indices.GetIndexRequest getIndexRequest = requestFactory.getIndexRequest(getIndexCoordinates());
+
+		return Mono.from(operations.executeWithIndicesClient(client -> client.getIndex(HttpHeaders.EMPTY, getIndexRequest)
```
sothawo, line 314: here the mapping should be done from Elasticsearch types to our data

sothawo (changes requested):
Thanks a lot for this PR, this looks pretty good. 
I left some minor remarks; what should be changed is mostly whate the `IndexInformation` class contains, I have added my comments there. 

And there should also be tests for the non-reactive code.

src/test/java/org/springframework/data/elasticsearch/core/index/IndexOperationTests.java
```diff
@@ -0,0 +1,62 @@
...
+
+		assertThat(indexInformation).isNotNull();
+		assertThat(indexInformation.getName()).isNotNull();
+		assertThat(indexInformation.getName()).isEqualTo(indexName);
+		assertThat(indexInformation.getMappings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+		assertThat(indexInformation.getSettings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+		assertThat(indexInformation.getAliases()).isInstanceOf(Set.class);
+
+	}
+}
```
sothawo, line 62: You should create an index with settings, mappings and alias data and check that it is really returned (you can use existing settings and mappings definition, must not really match the `Entity`):

```java
@SpringIntegrationTest
@ContextConfiguration(classes = { ElasticsearchRestTemplateConfiguration.class })
public class IndexOperationTests {
	public static final String INDEX_NAME = "random-index-name";
	@Autowired ElasticsearchOperations operations;

	@Test // #1646
	void shouldReturnInformationList() {
		String indexName = INDEX_NAME;
		IndexOperations indexOps = operations.indexOps(Entity.class);
		indexOps.create();
		indexOps.putMapping();
		indexOps.alias(new AliasActions(
				new AliasAction.Add(AliasActionParameters.builder().withIndices(INDEX_NAME).withAliases("alias").build())));
		List<IndexInformation> indexInformationList = indexOps.getInformation();
		assertThat(indexInformationList.size()).isEqualTo(1);

		IndexInformation indexInformation = indexInformationList.get(0);

// !!!!!!
// TODO: check the returned content here

		assertThat(indexInformation).isNotNull();
		assertThat(indexInformation.getName()).isNotNull();
		assertThat(indexInformation.getName()).isEqualTo(indexName);
		assertThat(indexInformation.getMappings())
				.isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
		assertThat(indexInformation.getSettings())
				.isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
		assertThat(indexInformation.getAliases()).isInstanceOf(Set.class);
	}

	@Data
	@Document(indexName = INDEX_NAME)
	@Setting(settingPath = "settings/test-settings.json")
	@Mapping(mappingPath = "mappings/test-mappings.json")
	private static class Entity {
		@Id String id;
	}
}
```

In addition, add another test class:

```java
@ContextConfiguration(classes = { ElasticsearchTemplateConfiguration.class })
public class IndexOperationTransportTests extends IndexOperationTests {}
```

This will inherit the test from the base class, but use the transport client to run the tests. If you then run this test, you will find the error in `RequestFactory`

src/test/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchTemplateIntegrationTests.java
```diff
@@ -1098,6 +1100,30 @@ void shouldReturnExplanationWhenRequested() {
...
+				.consumeNextWith(indexInformation -> {
+						assertThat(indexInformation.getName()).isNotNull();
+						assertThat(indexInformation.getName()).isInstanceOf(String.class);
+						assertThat(indexInformation.getMappings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+						assertThat(indexInformation.getSettings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+						assertThat(indexInformation.getAliases()).isInstanceOf(Set.class);
+				})
+				.verifyComplete();
+	}
+
```
sothawo, line 1126: see my comment for the non-reactive code in regards to checking the returned data

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
```diff
@@ -1841,5 +1834,78 @@ public Document fromSettingsResponse(GetSettingsResponse response, String indexN
...
+
+		return document;
+	}
+
+	public Document mappingsFromGetIndexResponse(org.elasticsearch.action.admin.indices.get.GetIndexResponse getIndexResponse, String indexName) {
+		Document document = Document.create();
+
+
+		if (getIndexResponse.getMappings().containsKey(indexName)) {
+			MappingMetadata mappings = getIndexResponse.getMappings().get(indexName).get(indexName);
```
sothawo, line 1898:
```suggestion
			MappingMetadata mappings = getIndexResponse.getMappings().get(indexName);
```
one too many

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
```diff
@@ -256,5 +259,29 @@ public boolean deleteTemplate(DeleteTemplateRequest deleteTemplateRequest) {
...
+					List<IndexInformation> indexInformationList = new ArrayList<>();
+
+					for (String indexName : getIndexResponse.getIndices()) {
+						Document settings = requestFactory.settingsFromGetIndexResponse(getIndexResponse, indexName);
+						Document mappings = requestFactory.mappingsFromGetIndexResponse(getIndexResponse, indexName);
+						Set<AliasData> indexAliases = aliases.get(indexName);
+
+						indexInformationList.add(IndexInformation.create(indexName, settings, mappings, indexAliases));
+					}
```
sothawo, line 280: Couldn't this whole loop be moved as well into the `RequestFactory`? Something like

```java
List<IndexInformation> indexInformationList = requestFactory.fromGetIndexResponse(getIndexResponse);
```

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
```diff
@@ -256,5 +259,29 @@ public boolean deleteTemplate(DeleteTemplateRequest deleteTemplateRequest) {
 				client -> client.indices().deleteTemplate(deleteIndexTemplateRequest, RequestOptions.DEFAULT).isAcknowledged());
 	}
 
+	@Override
+	public List<IndexInformation> getInformation() {
+		IndexCoordinates indexCoordinates = getIndexCoordinates();
+		GetIndexRequest request = requestFactory.getIndexRequest(indexCoordinates);
+
+		Map<String, Set<AliasData>> aliases = getAliasesForIndex(indexCoordinates.getIndexNames());
```
sothawo, line 267: this is executing a separate call to Elasticsearch for the aliases. But alias information is return in the GetIndex request, no need for a separate call here

src/main/java/org/springframework/data/elasticsearch/core/DefaultReactiveIndexOperations.java
```diff
@@ -304,6 +309,27 @@ public IndexCoordinates getIndexCoordinates() {
 		return (boundClass != null) ? getIndexCoordinatesFor(boundClass) : boundIndex;
 	}
 
+	@Override
+	public Flux<IndexInformation> getInformation() {
+		org.elasticsearch.client.indices.GetIndexRequest getIndexRequest = requestFactory.getIndexRequest(getIndexCoordinates());
+
+		return Mono.from(operations.executeWithIndicesClient(client -> client.getIndex(HttpHeaders.EMPTY, getIndexRequest)
+				.flatMap(getIndexResponse -> getAliasesForIndex(getIndexCoordinates().getIndexNames()).map(aliases -> {
```
sothawo, line 317: no separate call to get the alias information. This is returned in the GetIndex call

src/main/java/org/springframework/data/elasticsearch/core/DefaultReactiveIndexOperations.java
```diff
@@ -304,6 +309,27 @@ public IndexCoordinates getIndexCoordinates() {
...
+					List<IndexInformation> indexInformationList = new ArrayList<>();
+
+					for (String indexName : getIndexResponse.getIndices()) {
+						Document settings = requestFactory.settingsFromGetIndexResponse(getIndexResponse, indexName);
+						Document mappings = requestFactory.mappingsFromGetIndexResponse(getIndexResponse, indexName);
+						Set<AliasData> indexAliases = aliases.get(indexName);
+
+						indexInformationList.add(IndexInformation.create(indexName, settings, mappings, indexAliases));
+					}
```
sothawo, line 326: same comment as in the non-reactive code

src/main/java/org/springframework/data/elasticsearch/core/DefaultTransportIndexOperations.java
```diff
@@ -296,4 +301,27 @@ public boolean deleteTemplate(DeleteTemplateRequest deleteTemplateRequest) {
...
 	}
+
+	@Override
+	public List<IndexInformation> getInformation() {
+		GetIndexRequest getIndexRequest = new GetIndexRequest();
+		IndexCoordinates index = getIndexCoordinates();
+
+		getIndexRequest.indices(index.getIndexNames());
+
+	 	Map<String, Set<AliasData>>	aliases = getAliases(index.getIndexNames());
```
sothawo, line 312: no separate call for aliases, check my other comments

src/main/java/org/springframework/data/elasticsearch/core/DefaultTransportIndexOperations.java
```diff
@@ -296,4 +301,27 @@ public boolean deleteTemplate(DeleteTemplateRequest deleteTemplateRequest) {
...
+		GetIndexResponse getIndexResponse = client.admin().indices().getIndex(getIndexRequest).actionGet();
+		List<IndexInformation> indexInformationList = new ArrayList<>();
+
+		for (String indexName : getIndexResponse.getIndices()) {
+			Document settings = requestFactory.settingsFromGetIndexResponse(getIndexResponse, indexName);
+			Document mappings = requestFactory.mappingsFromGetIndexResponse(getIndexResponse, indexName);
+			Set<AliasData> indexAliases = aliases.get(indexName);
+
+			indexInformationList.add(IndexInformation.create(indexName, settings, mappings, indexAliases));
+		}
```
sothawo, line 323: see other comments

sothawo (changes requested):
We're getting there. One bug, some remarks about the tests.
And we must not issue an extra call to get alias information

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
gpopides: I also didn't really like writing the same thing 3 times.  But shouldn't request factory create only Request objects?
Is that a good place to add a function which returns a list of ```IndexInformation```?

src/test/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchTemplateIntegrationTests.java
gpopides: Is there a file/helper where i could put the following as a static method instead of copy/pasting the same lines?

```java

	private void assertIndexInformationIsCorrect(IndexInformation indexInformation, String indexName) {
		System.out.println(indexInformation.getMappings());
		assertThat(indexInformation.getSettings().get("index.number_of_shards")).isEqualTo("1");
		assertThat(indexInformation.getSettings().get("index.number_of_replicas")).isEqualTo("0");
		assertThat(indexInformation.getSettings().get("index.analysis.analyzer.emailAnalyzer.type")).isEqualTo("custom");

		assertThat(indexInformation.getMappings().get("properties.text.type")).isEqualTo("text");
		assertThat(indexInformation.getMappings().get("properties.text.analyzer")).isEqualTo("synonym_analyzer");

		assertThat(indexInformation.getName()).isEqualTo(indexName);
		assertThat(indexInformation.getMappings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
		assertThat(indexInformation.getSettings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
		assertThat(indexInformation.getAliases()).isInstanceOf(List.class);
	}
```

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
sothawo: fair point. The problem is that the reactive and non-reactive templates have no common base. Now that there are more than just one method for response data it might make sense to add a `ResponseConverter` class next to the `RequestFactory` that contains the methods for processing response data and that is used like the `Requestfactory`is.

src/main/java/org/springframework/data/elasticsearch/core/ResponseConverter.java
```diff
@@ -0,0 +1,66 @@
...
+		this.requestFactory = requestFactory;
+	}
+
+	public List<IndexInformation> indexInformationCollection(GetIndexResponse getIndexResponse) {
+		List<IndexInformation> indexInformationList = new ArrayList<>();
+
+
+		for (String indexName : getIndexResponse.getIndices()) {
+			Document settings = requestFactory.settingsFromGetIndexResponse(getIndexResponse, indexName);
+			Document mappings = requestFactory.mappingsFromGetIndexResponse(getIndexResponse, indexName);
```
sothawo, line 45: These functions should be moved from the `RequestFactory` to the `ResponseConverter`as well. This will also remove the dependency on the `RequestFactory`.

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
```diff
@@ -1841,5 +1834,72 @@ public Document fromSettingsResponse(GetSettingsResponse response, String indexN
...
+
+			for (String key : mappingsAsMap.keySet()) {
+				document.put(key, mappingsAsMap);
+			}
+		}
+
+		return document;
+	}
+
+
```
sothawo, line 1903: These should be moved to `ResponseConverter`.

src/test/java/org/springframework/data/elasticsearch/core/index/IndexOperationTransportTests.java
```diff
@@ -0,0 +1,75 @@
...
+
+import lombok.Data;
+
+/**
+ * @author George Popides
+ */
+
+@ContextConfiguration(classes = { ElasticsearchTemplateConfiguration.class })
+@SpringIntegrationTest
+public class IndexOperationTransportTests {
```
sothawo, line 29: It's not necessary to implement anything in this class. Just like I commented last time, all that's needed is a derived class with a different `@ContextConfiguration` annotation. When the tests are run, this derived class will set up the `ElasticsearchOperations` using the transport client and run the tests with this client. 
The `IndexOperationsTests` class just uses the methods from `ElasticsearchOperations` not caring which implementation it gets.

src/test/java/org/springframework/data/elasticsearch/core/index/IndexOperationTests.java
```diff
@@ -0,0 +1,85 @@
...
+		assertThat(indexInformation.getSettings().get("index.number_of_shards")).isEqualTo("1");
+		assertThat(indexInformation.getSettings().get("index.number_of_replicas")).isEqualTo("0");
+		assertThat(indexInformation.getSettings().get("index.analysis.analyzer.emailAnalyzer.type")).isEqualTo("custom");
+
+		assertThat(indexInformation.getMappings()).containsKey("properties");
+
+		assertThat(indexInformation.getName()).isEqualTo("test-index-rest-information-list");
+		assertThat(indexInformation.getMappings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+		assertThat(indexInformation.getSettings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+		assertThat(indexInformation.getAliases()).isInstanceOf(List.class);
```
sothawo, line 74: please check the content of the returned alias information

src/main/java/org/springframework/data/elasticsearch/core/DefaultIndexOperations.java
```diff
@@ -60,22 +62,26 @@
...
  * @author Sascha Woo
+ * @author George Popides
  * @since 4.0
  */
 class DefaultIndexOperations extends AbstractDefaultIndexOperations implements IndexOperations {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(DefaultIndexOperations.class);
 
 	private final ElasticsearchRestTemplate restTemplate;
+	private final ResponseConverter responseConverter;
```
sothawo, line 73: this can be moved to the parent `AbstractDefaultIndexOperations` and be `protected`

src/main/java/org/springframework/data/elasticsearch/core/DefaultTransportIndexOperations.java
```diff
@@ -66,24 +69,29 @@
...
  * @author Peter-Josef Meisch
  * @author Sascha Woo
+ * @author George Popides
  * @since 4.0
  */
 class DefaultTransportIndexOperations extends AbstractDefaultIndexOperations implements IndexOperations {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(DefaultTransportIndexOperations.class);
 
+	private final ResponseConverter responseConverter;
```
sothawo, line 79: move to the base class

src/test/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchTemplateIntegrationTests.java
```diff
@@ -1098,6 +1101,35 @@ void shouldReturnExplanationWhenRequested() {
...
+					assertThat(indexInformation.getSettings().get("index.number_of_shards")).isEqualTo("1");
+					assertThat(indexInformation.getSettings().get("index.number_of_replicas")).isEqualTo("0");
+					assertThat(indexInformation.getSettings().get("index.analysis.analyzer.emailAnalyzer.type")).isEqualTo("custom");
+
+					assertThat(indexInformation.getMappings()).containsKey("properties");
+
+					assertThat(indexInformation.getName()).isEqualTo(indexName);
+					assertThat(indexInformation.getMappings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+					assertThat(indexInformation.getSettings()).isInstanceOf(org.springframework.data.elasticsearch.core.document.Document.class);
+					assertThat(indexInformation.getAliases()).isInstanceOf(List.class);
```
sothawo, line 1128: here, check the content of the returned alias data

sothawo (changes requested):
a couple of refactorings, then this should be fine

src/test/java/org/springframework/data/elasticsearch/core/index/IndexOperationTransportTests.java
gpopides: I tried to do it in this way, but the the tests when were run all together were failing with the following exception:
``` Elasticsearch exception [type=resource_already_exists_exception, ...``` for ```test-index-rest-information-list``` index.

I removed one of the entities and left just one with index name: ```test-index-information-list``` but the test still failed.
I have tried to remove ```indexOps.create()``` from one of the methods, but it didn't work either.

If the test is run alone, it passes.  So that's why i didn't extend the ```IndexOperationTests``` class in previous commit.

Comment (sothawo) [reply src/test/java/org/springframework/data/elasticsearch/core/index/IndexOperationTransportTests.java:29]: forgot that. You need a method
```java
    @BeforeEach
    void setUp() {
        operations.indexOps(EntityWithSettingsAndMappingsRest.class).delete();
    }
```
in the base class that deletes the index on test start when it exists.

src/test/java/org/springframework/data/elasticsearch/core/index/IndexOperationTransportTests.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+@ContextConfiguration(classes = { ElasticsearchTemplateConfiguration.class })
+public class IndexOperationTransportTests extends IndexOperationTests {
+	@Autowired
+	protected ElasticsearchOperations operations;
+
+	@Test // #1646
+	@DisplayName("should return info of all indices using TRANSPORT template with no aliases")
+	public void shouldReturnInformationListOfAllIndicesNoAliases() throws JSONException {
```
sothawo, line 30: this method is not needed. It does the same as the one in the base class. The purpose of the `IndexOperationTransportTests` class is to run exactly the same tests as in `IndexOperationTests` just having a different implementation of `ElasticsearchOperations`.

src/test/java/org/springframework/data/elasticsearch/core/index/IndexOperationTests.java
```diff
@@ -0,0 +1,89 @@
...
+	@Autowired
+	protected ElasticsearchOperations operations;
+
+	@BeforeEach
+	void setUp() {
+		operations.indexOps(EntityWithSettingsAndMappings.class).delete();
+	}
+
+	@Test // #1646
+	@DisplayName("should return info of all indices using REST template without aliases")
```
sothawo, line 56: why without aliases? The test should cover that alias information is returned as well.

src/main/java/org/springframework/data/elasticsearch/core/mapping/IndexInformation.java
```diff
@@ -0,0 +1,78 @@
...
+ * @since 4.2
+ */
+public class IndexInformation {
+	private final String name;
+	@Nullable
+	private final Document settings;
+	@Nullable
+	private final Document mappings;
+	@Nullable
+	private final List<AliasMetadata> aliases;
```
sothawo, line 38: I just now noticed that here we use `AliasMetadata` from Elasticsearch. We should use `org.springframework.data.elasticsearch.core.index.AliasData`, the same class that is used in the return values of `org.springframework.data.elasticsearch.core.AbstractDefaultIndexOperations.getAliasesForIndex()`

src/main/java/org/springframework/data/elasticsearch/core/ResponseConverter.java
```diff
@@ -0,0 +1,140 @@
...
+	// region alias
+
+	public List<IndexInformation> indexInformationCollection(GetIndexResponse getIndexResponse) {
+		List<IndexInformation> indexInformationList = new ArrayList<>();
+
+
+		for (String indexName : getIndexResponse.getIndices()) {
+			Document settings = settingsFromGetIndexResponse(getIndexResponse, indexName);
+			Document mappings = mappingsFromGetIndexResponse(getIndexResponse, indexName);
+			List<AliasMetadata> aliases = getIndexResponse.getAliases().get(indexName) != null
```
sothawo, line 49: Use `org.springframework.data.elasticsearch.core.index.AliasData` in these functions

src/test/java/org/springframework/data/elasticsearch/core/ReactiveElasticsearchTemplateIntegrationTests.java
```diff
@@ -1098,6 +1101,36 @@ void shouldReturnExplanationWhenRequested() {
...
 				.verifyComplete();
 	}
+
+
+	@Test // #1646
+	@DisplayName("should return info of all indices using reactive template")
+	void shouldReturnInformationListOfAllIndices() {
+		String indexName = "test-index-reactive-information-list";
+		ReactiveIndexOperations indexOps = template.indexOps(EntityWithSettingsAndMappingsReactive.class);
+
```
sothawo, line 1111: set and test alias information as well

sothawo (changes requested):
didn't notice before, but the returned alias information should be a `org.springframework.data.elasticsearch.core.index.AliasData`, not the Elasticsearch class

sothawo:
and please rebase your changes on the current master branch

src/main/java/org/springframework/data/elasticsearch/core/mapping/IndexInformation.java
gpopides: I'm searching for the way to conver ```AliasMetadata``` to ```AliasData``` but something looks off.

The ```filter``` atribute of ```AliasData``` is ```Document``` which essentialy is a ```Map```, but the ```filter``` attribute of ```AliasMetadata``` is ```CompressedXContent```.  So how do i convert it to a Document?

sothawo: there already is `org.springframework.data.elasticsearch.core.RequestFactory.convertAliasMetadata()` - which probably could be moved to `ResponseConverter` as well

PR #1728, opened 2021-03-15
Allow multiple date formats for date fields

### LINKED ISSUE #1727: Allow multiple date formats for date fields
At the moment it is not possible to specify multiple date formats for date fields, but ES supports it. The default is `strict_date_optional_time||epoch_millis`.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html

xhaggi:
Closes #1727 

- [x] There is a ticket in the bug tracker for the project in our [issue tracker](https://github.com/spring-projects/spring-data-elasticsearch/issues).

### DIFF
--- a/src/main/asciidoc/reference/elasticsearch-object-mapping.adoc
+++ b/src/main/asciidoc/reference/elasticsearch-object-mapping.adoc
@@ -56,16 +56,18 @@ Default value is _EXTERNAL_.
 Constructor arguments are mapped by name to the key values in the retrieved Document.
 * `@Field`: Applied at the field level and defines properties of the field, most of the attributes map to the respective https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html[Elasticsearch Mapping] definitions (the following list is not complete, check the annotation Javadoc for a complete reference):
 ** `name`: The name of the field as it will be represented in the Elasticsearch document, if not set, the Java field name is used.
-** `type`: the field type, can be one of _Text, Keyword, Long, Integer, Short, Byte, Double, Float, Half_Float, Scaled_Float, Date, Date_Nanos, Boolean, Binary, Integer_Range, Float_Range, Long_Range, Double_Range, Date_Range, Ip_Range, Object, Nested, Ip, TokenCount, Percolator, Flattened, Search_As_You_Type_.
+** `type`: The field type, can be one of _Text, Keyword, Long, Integer, Short, Byte, Double, Float, Half_Float, Scaled_Float, Date, Date_Nanos, Boolean, Binary, Integer_Range, Float_Range, Long_Range, Double_Range, Date_Range, Ip_Range, Object, Nested, Ip, TokenCount, Percolator, Flattened, Search_As_You_Type_.
 See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html[Elasticsearch Mapping Types]
-** `format` and `pattern` definitions for the _Date_ type.
+** `format`: One or more built-in formats, default value is _strict_date_optional_time_ and _epoch_millis_.
+See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#built-in-date-formats[Elasticsearch Built In Formats]
+** `pattern`: One or more custom date formats. NOTE: If you want to use only custom date formats, you must set the `format` property to empty `{}`.
+See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#custom-date-formats[Elasticsearch Custom Date Formats]
 ** `store`: Flag whether the original field value should be store in Elasticsearch, default value is _false_.
 ** `analyzer`, `searchAnalyzer`, `normalizer` for specifying custom analyzers and normalizer.
-* `@GeoPoint`: marks a field as _geo_point_ datatype.
+* `@GeoPoint`: Marks a field as _geo_point_ datatype.
 Can be omitted if the field is an instance of the `GeoPoint` class.
 
-NOTE: Properties that derive from `TemporalAccessor` or are of type `java.util.Date` must either have a `@Field` annotation of type `FieldType.Date` and a
-format different from `DateFormat.none` or a custom converter must be registered for this type. +
+NOTE: Properties that derive from `TemporalAccessor` or are of type `java.util.Date` must either have a `@Field` annotation of type `FieldType.Date`.
 If you are using a custom date format, you need to use _uuuu_ for the year instead of _yyyy_.
 This is due to a https://www.elastic.co/guide/en/elasticsearch/reference/current/migrate-to-java-time.html#java-time-migration-incompatible-date-formats[change in Elasticsearch 7].
 
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/DateFormat.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/DateFormat.java
@@ -23,9 +23,21 @@ package org.springframework.data.elasticsearch.annotations;
  * @author Jakub Vavrik
  * @author Tim te Beek
  * @author Peter-Josef Meisch
+ * @author Sascha Woo
  */
 public enum DateFormat {
+	/**
+	 * @deprecated since 4.2, will be removed in a future version. Use <code>format = {}</code> to disable built-in date
+	 *             formats in the @Field annotation.
+	 */
+	@Deprecated
 	none(""), //
+	/**
+	 * @deprecated since 4.2, will be removed in a future version.It is no longer required for using a custom date format
+	 *             pattern. If you want to use only a custom date format pattern, you must set the <code>format</code>
+	 *             property to empty <code>{}</code>.
+	 */
+	@Deprecated
 	custom(""), //
 	basic_date("uuuuMMdd"), //
 	basic_date_time("uuuuMMdd'T'HHmmss.SSSXXX"), //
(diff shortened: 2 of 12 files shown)


src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -65,7 +66,7 @@
 
 	boolean index() default true;
 
-	DateFormat format() default DateFormat.none;
+	DateFormat[] format() default { DateFormat.date_optional_time, DateFormat.epoch_millis };
```
sothawo, line 69: comment only: no breaking change here, because `@Field(format=DateFormat.none)` still works. Sometimes Java surprieses me.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -154,51 +155,79 @@ private void initDateConverter() {
...
+									return dateConverter.parse(s, (Class<? extends TemporalAccessor>) actualType);
+								} catch (Exception e) {
+								}
+							}
+						} else { // must be date
+							for (ElasticsearchDateConverter dateConverter : dateConverters) {
+								try {
+									return dateConverter.parse(s);
+								} catch (Exception e) {
+								}
```
sothawo, line 220: we should at least log the exception here

src/test/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchDateConverterUnitTests.java
```diff
@@ -33,7 +34,6 @@
 	void shouldCreateConvertersForAllKnownFormats(DateFormat dateFormat) {
 
 		switch (dateFormat) {
-			case none:
```
sothawo, line 36: see my remark in the enum class

sothawo (changes requested):
long time no see. 
Basically one thing: deprecate `none` instead of removing it, this is a breaking change

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -65,7 +66,7 @@
...
 
 	String pattern() default "";
```
xhaggi, line 71: @sothawo not sure if we want to support multiple custom format patterns as well.

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
xhaggi: If no converter can parse the date, we throw a `RuntimeException` at the very end.

src/main/java/org/springframework/data/elasticsearch/annotations/DateFormat.java
```diff
@@ -23,8 +23,13 @@
  * @author Jakub Vavrik
  * @author Tim te Beek
  * @author Peter-Josef Meisch
+ * @author Sascha Woo
  */
 public enum DateFormat {
+	/**
+	 * @deprecated will be removed in a future version
```
sothawo, line 30:
```suggestion
	 * @deprecated since 4.2, will be removed in a future version, use multiple formats in the @Field annotation.
```

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
sothawo: why not?

src/test/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchDateConverterUnitTests.java
```diff
@@ -23,6 +23,7 @@
 /**
  * @author Peter-Josef Meisch
  * @author Tim te Beek
+ * @author Sascha Woo
```
sothawo, line 26: any other changes in this file? 🤔

sothawo (changes requested):
can you please adapt the documentation in https://github.com/spring-projects/spring-data-elasticsearch/blob/master/src/main/asciidoc/reference/elasticsearch-object-mapping.adoc, lines 67++ to this change?
Otherwise only cosmetic changes

src/test/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchDateConverterUnitTests.java
xhaggi: No, removed.

src/main/java/org/springframework/data/elasticsearch/annotations/DateFormat.java
xhaggi: Done.

xhaggi:
@sothawo I have addressed all of your comments and force-pushed the changes.

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
xhaggi: Ok, I also added support for multiple custom formats.

src/main/asciidoc/reference/elasticsearch-object-mapping.adoc
```diff
@@ -64,8 +64,7 @@ See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-type
 * `@GeoPoint`: marks a field as _geo_point_ datatype.
 Can be omitted if the field is an instance of the `GeoPoint` class.
 
-NOTE: Properties that derive from `TemporalAccessor` or are of type `java.util.Date` must either have a `@Field` annotation of type `FieldType.Date` and a
-format different from `DateFormat.none` or a custom converter must be registered for this type. +
+NOTE: Properties that derive from `TemporalAccessor` or are of type `java.util.Date` must either have a `@Field` annotation of type `FieldType.Date`.
```
sothawo, line 67: I think we should add a new section here about DateFormats and explain how multiple formats are mapped to Elasticsearch and also describe that for a custom format there may be multiple patterns; best would be a sample `@Field` annotation and the mapping that is created in Elasticsearch.

src/main/java/org/springframework/data/elasticsearch/core/index/MappingParameters.java
```diff
@@ -226,8 +229,22 @@ public void writeTypeAndParametersTo(XContentBuilder builder) throws IOException
...
+				for (int i = 0; i < dateFormats.length; i++) {
+					DateFormat dateFormat = dateFormats[i];
+					if (dateFormat == DateFormat.custom) {
+						for (int ii = 0; ii < dateFormatPatterns.length; ii++) {
+							formats.add(dateFormatPatterns[ii]);
+						}
+					} else {
+						formats.add(dateFormat.toString());
+					}
+				}
```
sothawo, line 245:
```suggestion
                for (DateFormat dateFormat : dateFormats) {
                    
                    if (dateFormat == DateFormat.custom) {
                        Collections.addAll(formats, dateFormatPatterns);
                    } else {
                        formats.add(dateFormat.toString());
                    }
                }
```
a little shorter

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -154,51 +155,86 @@ private void initDateConverter() {
...
+					@SuppressWarnings("unchecked")
+					@Override
+					public Object read(String s) {
+						if (isTemporalAccessor) {
+							for (ElasticsearchDateConverter dateConverter : dateConverters) {
+								try {
+									return dateConverter.parse(s, (Class<? extends TemporalAccessor>) actualType);
+								} catch (Exception e) {
+								}
+							}
```
sothawo, line 221: after the loop, if every registered converter fails, we should throw an error that none of the defined date formats could be used to read the value

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
```diff
@@ -154,51 +155,86 @@ private void initDateConverter() {
...
+						} else { // must be date
+							for (ElasticsearchDateConverter dateConverter : dateConverters) {
+								try {
+									return dateConverter.parse(s);
+								} catch (Exception e) {
+								}
+							}
```
sothawo, line 228: same as above, when all converters fail, we must error

sothawo (changes requested):
some remarks, and: there are no tests that test that the correct mapping is built for multiple formats

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentProperty.java
xhaggi: This is done for both at the end of the method. Why should we do it twice?

src/main/java/org/springframework/data/elasticsearch/core/index/MappingParameters.java
xhaggi: Oh yes, that's true ;)

src/main/asciidoc/reference/elasticsearch-object-mapping.adoc
xhaggi: I have changed the documentation so that it is clear what values can be set to `format` and `pattern` and I have linked the ES documentation pages.

xhaggi:
@sothawo I've done some more changes that in my opinion makes things a bit easier. Now it is no longer necessary to set `DateFormat.custom` to use Custom Date Formats. If one or more values are set in `pattern` they are used without checking for `DateFormat.custom`. This way `DateFormat.custom` is deprecated and can be removed in a future version, as well as `none`. I have adjusted the documentation so that it is clear that you must set the `format` property to empty `{}` if you want to use only custom date formats.

sothawo (approved):
I'll change the thrown exception and add more documentation and test in a following polishing commit.

xhaggi:
👍 thank you

PR #1856, opened 2021-07-01
Support collection parameters in @Query methods

### LINKED ISSUE #1858: Collection parameters for @Query-annotated methods get escaped wrongly
The `@Query` annotation breaks when providing a collection parameter - it calls `toString()` on the parameter and escapes the result, which makes the list look like this `[\"some\",\"params\"]`, causing a 400 from Elasticsearch

herder:
This adds a test and fix so that collection parameters are supported and quoted correctly, at least for `Collection<String>`.

Previously the `StringQueryUtil` class would just call `.toString()` on the parameter and escape the `"` characters in the result, which would cause a syntax error in the query string.

This changes so that the class checks if the parameter is a `Collection` and if so starts a Json array and adds elements to it by calling `convert()` on each element.

Closes #1858 

- [x]  You have read the Spring Data contribution guidelines.
- [x]  There is a ticket in the bug tracker for the project in our issue tracker.
- [x]  You use the code formatters provided here and have them applied to your changes. Don’t submit any formatting related changes.
- [x]  You submit test cases (unit or integration tests) that back your changes.
- [x]  You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/repository/support/StringQueryUtil.java
+++ b/src/main/java/org/springframework/data/elasticsearch/repository/support/StringQueryUtil.java
@@ -15,8 +15,10 @@
  */
 package org.springframework.data.elasticsearch.repository.support;
 
+import java.util.Collection;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.springframework.core.convert.support.GenericConversionService;
 import org.springframework.data.repository.query.ParameterAccessor;
@@ -24,6 +26,7 @@ import org.springframework.util.NumberUtils;
 
 /**
  * @author Peter-Josef Meisch
+ * @author Niklas Herder
  */
 final public class StringQueryUtil {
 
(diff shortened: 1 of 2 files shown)


sothawo (commented):
Thanks a lot for this PR. This looks good, I just added a comment/question on this in the code

src/main/java/org/springframework/data/elasticsearch/repository/support/StringQueryUtil.java
```diff
@@ -53,6 +56,22 @@ private static String getParameterWithIndex(ParameterAccessor accessor, int inde
...
+		return parameterValue;
+
+	}
+
+	private static String convert(Object parameter) {
+		if (Collection.class.isAssignableFrom(parameter.getClass())) {
+			Collection<?> collectionParam = (Collection<?>) parameter;
+			StringBuilder sb = new StringBuilder("[");
+			sb.append(collectionParam.stream().map(o -> "\"" + convert(o) + "\"").collect(Collectors.joining(",")));
+			sb.append("]");
```
sothawo, line 71: This works, but what do you think about
```java
			StringBuilder sb = new StringBuilder("[");
			sb.append(collectionParam.stream().map(o -> {
				if (o instanceof String) {
					String s = (String) o;
					return '"' + convert(o) + '"';
				} else {
					return convert(o);
				}
			}).collect(Collectors.joining(",")));
			sb.append(']');
```
Might be better to only wrap Strings in quotes.

herder: Yes, that makes sense. I'll add a test and incorporate your change.

herder: Fixed in 8b154dfd , @sothawo

PR #1872, opened 2021-07-15
Move dynamic mapping parameter configuration to @Document and @Field

### LINKED ISSUE #1871: Move dynamic mapping parameter configuration to @Document and @Field
At the moment you have to add a separate annotation `@DynamicMapping` to the entity or fields of type `Object` and `Nested` to configure the dynamic mapping parameter. This seems a bit too complicated to me, since the other mapping parameters like `ignore_above` or `doc_values` are part of the `@Field` annotation. We should add `dynamic` as part of the `@Document` and `@Field` annotation, deprecate the old implementation and switch to the new one.

xhaggi:

Closes #1871

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/index/MappingBuilder.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/index/MappingBuilder.java
@@ -197,7 +197,9 @@ public class MappingBuilder {
 			}
 		}
 
-		if (dynamicMapping != null) {
+		if (entity != null && entity.dynamic() != Dynamic.INHERIT) {
+			builder.field(TYPE_DYNAMIC, entity.dynamic().name().toLowerCase());
+		} else if (dynamicMapping != null) {
 			builder.field(TYPE_DYNAMIC, dynamicMapping.value().name().toLowerCase());
 		}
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -102,6 +104,7 @@ public SimpleElasticsearchPersistentEntity(TypeInformation<T> typeInformation,
 			this.indexName = document.indexName();
 			this.versionType = document.versionType();
 			this.createIndexAndMapping = document.createIndex();
+			this.dynamic = document.dynamic();
 		}
```
sothawo, line 109:
```suggestion
			this.dynamic = document.dynamic();
		} else {
                       this.dynamic = Dynamic.INHERIT;
               }
```

see above

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -73,6 +74,7 @@
 	private @Nullable ElasticsearchPersistentProperty joinFieldProperty;
 	private @Nullable VersionType versionType;
 	private boolean createIndexAndMapping;
+	private Dynamic dynamic = Dynamic.INHERIT;
```
sothawo, line 77:
```suggestion
	private final Dynamic dynamic = Dynamic;
```

I'd prefer a `final`property

sothawo (changes requested):
some minor remarks, a failing test class

src/main/java/org/springframework/data/elasticsearch/core/index/MappingBuilder.java
```diff
@@ -197,7 +197,12 @@ private void mapEntity(XContentBuilder builder, @Nullable ElasticsearchPersisten
 			}
 		}
 
-		if (dynamicMapping != null) {
+		if (isRootObject && entity != null && entity.dynamic() != Dynamic.INHERIT) {
+			builder.field(TYPE_DYNAMIC, entity.dynamic().name().toLowerCase());
+		} else if (nestedOrObjectField && parentFieldAnnotation != null
+				&& parentFieldAnnotation.dynamic() != Dynamic.INHERIT) {
+			builder.field(TYPE_DYNAMIC, parentFieldAnnotation.dynamic().name().toLowerCase());
```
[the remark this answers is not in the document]
xhaggi: The method is called twice, for the root and for a nested entity and I thought it was necessary, but it's not. To make both work, I simply removed the "isRootObject" from the "if" statement.

sothawo (approved):

PR #1873, opened 2021-07-15
fix uri encode bug when url path start with '/' #1870

### LINKED ISSUE #1870: fix uri encode bug when url path start with '/'
Please take a look at this fix in original elasticsearch client code:

https://github.com/elastic/elasticsearch/pull/34436

please update spring data's RequestConverters#encodePart to reflect the fix in this issue, since the implementation in spring-data  seems to be a copy of an older (pre-fix) version of encodePart, specifically building the URI.
Is there a strong reason to have this duplicated in spring-data, couldn't it use same code from original ES high level client?
Thanks

frnandu:
Closes #1870


sothawo:
Thanks for this PR!

PR #1925, opened 2021-09-09
Change mapping of connectionRequestTimeout to ConnPool leaseTimeout

nichines:
Currently the RestClients.create() method sets the underlying RequestConfig connectTimeout and connectionRequestTimeout properties to the value of the ClientConfiguration connectTimeout property. 

The connectionRequestTimeout is used as the leaseTimeout parameter when calling the AbstractNIOConnPool lease method, which can lead to a "TimeoutException: Connection lease request time out" on an establised connection.  

The correct behavior is to use the sockerTimeout property for the leastTimeout parameter.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/client/RestClients.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/RestClients.java
@@ -55,6 +55,7 @@ import org.springframework.util.Assert;
  * @author Huw Ayling-Miller
  * @author Henrique Amaral
  * @author Peter-Josef Meisch
+ * @author Nic Hines
  * @since 3.2
  */
 public final class RestClients {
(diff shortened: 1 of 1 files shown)


sothawo:
Thank you very much you for finding and fixing this. 
Please create an issue first the next time before the PR.

src/main/java/org/springframework/data/elasticsearch/client/RestClients.java
```diff
@@ -105,15 +106,14 @@ public static ElasticsearchRestClient create(ClientConfiguration clientConfigura
...
-
 				requestConfigBuilder.setConnectTimeout(Math.toIntExact(connectTimeout.toMillis()));
-				requestConfigBuilder.setConnectionRequestTimeout(Math.toIntExact(connectTimeout.toMillis()));
 			}
 
 			Duration timeout = clientConfiguration.getSocketTimeout();
 
 			if (!timeout.isNegative()) {
 				requestConfigBuilder.setSocketTimeout(Math.toIntExact(timeout.toMillis()));
+				requestConfigBuilder.setConnectionRequestTimeout(Math.toIntExact(connectTimeout.toMillis()));
```
sothawo, line 116: just at a second look I saw that you moved the line into the _if_ for the socketTimeout, but still using the configured value for the connectTimeout. Shouldn't do PRs in the morning before work. I'll fix that.

PR #1998, opened 2021-11-16
Fix IndexOutOfBoundsException when try to map inner hits with no results returned

### LINKED ISSUE #1997: IndexOutOfBoundsException when try to map inner hits with no results returned
Configuring `inner_hits` with a size of 0 will result in an `IndexOutOfBoundsException` when mapping inner hits.

```
WARN  o.s.d.e.core.SearchHitMapping - Could not map inner_hits
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at org.springframework.data.elasticsearch.core.SearchHitsImpl.getSearchHit(SearchHitsImpl.java:97)
	at org.springframework.data.elasticsearch.core.SearchHitMapping.mapInnerDocuments(SearchHitMapping.java:181)
	at org.springframework.data.elasticsearch.core.SearchHitMapping.mapInnerHits(SearchHitMapping.java:158)
	at org.springframework.data.elasticsearch.core.SearchHitMapping.mapHit(SearchHitMapping.java:115)
	at org.springframework.data.elasticsearch.core.SearchHitMapping.mapHitsFromResponse(SearchHitMapping.java:95)
	at org.springframework.data.elasticsearch.core.SearchHitMapping.mapHits(SearchHitMapping.java:71)
	at org.springframework.data.elasticsearch.core.AbstractElasticsearchTemplate$ReadSearchDocumentResponseCallback.doWith(AbstractElasticsearchTemplate.java:875)
	at org.springframework.data.elasticsearch.core.AbstractElasticsearchTemplate$ReadSearchDocumentResponseCallback.doWith(AbstractElasticsearchTemplate.java:860)
	at org.springframework.data.elasticsearch.core.ElasticsearchRestTemplate.search(ElasticsearchRestTemplate.java:299)
	at org.springframework.data.elasticsearch.core.AbstractElasticsearchTemplate.search(AbstractElasticsearchTemplate.java:513)
	at ch.x28.jobportal.core.searchindex.SearchIndexRepositoryImpl.findAll(SearchIndexRepositoryImpl.java:56)
```

xhaggi:
Fixes #1997 

Closes #issue-number

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/SearchHitMapping.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/SearchHitMapping.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.data.elasticsearch.UncategorizedElasticsearchException;
 import org.springframework.data.elasticsearch.backend.elasticsearch7.document.SearchDocumentResponse;
 import org.springframework.data.elasticsearch.core.convert.ElasticsearchConverter;
 import org.springframework.data.elasticsearch.core.document.Document;
@@ -45,6 +46,7 @@ import org.springframework.util.Assert;
  * @author Mark Paluch
  * @author Roman Puchkovskiy
  * @author Matt Gilene
+ * @author Sascha Woo
  * @since 4.0
  */
 public class SearchHitMapping<T> {
@@ -194,7 +196,7 @@ public class SearchHitMapping<T> {
 	 */
 	private SearchHits<?> mapInnerDocuments(SearchHits<SearchDocument> searchHits, Class<T> type) {
 
-		if (searchHits.getTotalHits() == 0) {
+		if (searchHits.isEmpty()) {
 			return searchHits;
 		}
 
@@ -239,7 +241,7 @@ public class SearchHitMapping<T> {
 						searchHits.getSuggest());
 			}
 		} catch (Exception e) {
-			LOGGER.warn("Could not map inner_hits", e);
+			throw new UncategorizedElasticsearchException("Unable to convert inner hits.", e);
 		}
 
 		return searchHits;
--- a/src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchTemplateTests.java
@@ -2787,6 +2787,33 @@ public abstract class ElasticsearchTemplateTests {
 		assertThat(searchHits.getSearchHit(1).getInnerHits("innerHits").getTotalHits()).isEqualTo(1);
 	}
 
+	@Test // #1997
+	@DisplayName("should return document with inner hits size zero")
+	void shouldReturnDocumentWithInnerHitsSizeZero() {
+
+		// given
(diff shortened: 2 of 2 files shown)


src/main/java/org/springframework/data/elasticsearch/core/SearchHitMapping.java
```diff
@@ -194,7 +195,7 @@ public void mapHitsInCompletionSuggestion(@Nullable Suggest suggest) {
 	 */
 	private SearchHits<?> mapInnerDocuments(SearchHits<SearchDocument> searchHits, Class<T> type) {
 
-		if (searchHits.getTotalHits() == 0) {
+		if (searchHits.isEmpty()) {
```
sothawo, line 198: Do you have a test case showing the error with the old code? `SearchHits.getTotalHits()` returns a primitive `long` value. How does this produce an `IndexOutOfBoundsException`? It might be a `NaN` but then we should rather fix the setting of the `totalHits`property.

sothawo (changes requested):

src/main/java/org/springframework/data/elasticsearch/core/SearchHitMapping.java
xhaggi: The `IndexOutOfBoundsException` is thrown in line 203 when accessing `searchHits.getSearchHit(0)`. You can tell Elastic not to return inner hits by setting the `size` in `inner_hits` to 0. This way you can keep track of total hits without returning any results.

xhaggi: By the way, it's not so easy to add a test case for this, because the exceptions are caught here and only a warning is logged. I am not sure if we should change that as well.

sothawo: ok for the change. 

The exception should probably not be swallowed, it rather should be wrapped in a `UncategorizedElasticsearchException`.

As for the test: this could be done with a method in `org.springframework.data.elasticsearch.core.ElasticsearchTemplateTests`, insert a document, then do a search with the `inner_hits` set to zero which should then produce the exception with the old code.

AntonLGVS: **searchHits.getTotalHits() == searchHits.getSearchHits().size()** ?
**_searchHits.getTotalHits()_** - number of found.
**_searchHits.getSearchHits()_** - number of received for current page.
They may not be equal. Or am I wrong?

maybe: 
```
if (!searchHits.hasSearchHits()) {
```

and **searchHits.getSearchHit(0)** not throw _IndexOutOfBoundsException_

xhaggi: @sothawo everything has now been addressed and force-pushed.

xhaggi:
@sothawo a backport to 4.2 and 4.3 would be great to get rid of the warning. Let me know if you want me to do that.

PR #2007, opened 2021-11-22
Fix flaky test in MappingElasticsearchConverterUnitTests

CharlesZKQ:
In `org.springframework.data.elasticsearch.core.convert.MappingElasticsearchConverterUnitTests`, the test `shouldMapObjectToJsonString()` is flaky due to the non-deterministic property of `java.lang.Class.getMethods()` (please refer [document](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getMethods--)), which is internally invoked by `mapObject()`. I fixed it by changing `JSONSTRING` to a `Set`, so that either one of the possible `jsonResult` will pass the test.

### DIFF
--- a/src/test/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverterUnitTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/core/convert/MappingElasticsearchConverterUnitTests.java
@@ -19,7 +19,6 @@ import static java.util.Collections.*;
 import static org.assertj.core.api.Assertions.*;
 import static org.skyscreamer.jsonassert.JSONAssert.*;
 
-import java.io.Serializable;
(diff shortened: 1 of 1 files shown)


sothawo:
Why is this test flaky? Did you encounter a failing test?

`buildPropertyMapping()` does not call `java.lang.Class.getMethods()`.  

And even if the values of the JSON objects were in different order, this would be of no concern to the `org.skyscreamer.jsonassert.JSONAssert#assertEquals(java.lang.String, java.lang.String, boolean)` method even in strict mode. 

If you have an issue with a test that does not run correctly, then please create an issue describing that problem with an reproducible example. But posting this kind of solution to obviously not existing problems as 15 PRs in 6 repositories does not help.

PR #2017, opened 2021-11-30
Add reactive SearchHits.

### LINKED ISSUE #2015: Totalhits support for ReactiveSearchOperations
Hello,

I need to have total count while searching and would like to avoid making two searches in Elasticsearch.
For non-reactive version of `SearchOperations` I could use the following method:

`<T> SearchHits<T> search(Query query, Class<T> clazz);`

But I miss the corresponding method in `ReactiveSearchOperations`.  It would be very convenient  to have something similar in `ReactiveSearchOperations` like:

`<T> ReactiveSearchHits<T> search(Query query, Class<T> clazz);`

where the `ReactiveSearchHits` is a reactive version of `SearchHits` returning `Flux` instead of `List`.

Is this possible to deliver such an API or are there any alternative ways for doing that in the reactive world?

I am using Spring Boot 2.5.6 with Spring Data ElasticSearch 4.0.4. Thank you in advance for your help!

sothawo:
Closes #2015

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/ReactiveSearchHitSupport.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/ReactiveSearchHitSupport.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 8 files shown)


src/main/java/org/springframework/data/elasticsearch/core/ReactiveSearchHitsImpl.java
```diff
@@ -0,0 +1,80 @@
...
+package org.springframework.data.elasticsearch.core;
+
+import reactor.core.publisher.Flux;
+
+import org.springframework.data.elasticsearch.core.suggest.response.Suggest;
+import org.springframework.lang.Nullable;
+
+/**
+ * @author Peter-Josef Meisch
+ */
```
sothawo, line 25: @since 4.4

src/main/java/org/springframework/data/elasticsearch/core/ReactiveSearchHitSupport.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import org.springframework.util.Assert;
+
+/**
+ * @author Peter-Josef Meisch
+ */
```
sothawo, line 22: @since 4.4

Issue #2105, opened 2022-03-04
Support searchAfter in NativeSearchQueryBuilder 


owen-q:
NativeSearchQuery has searchAfter options but it's builder does not have searchAfter interfaces

PR #2106, opened 2022-03-04
Add searchAfter interfaces to NativeSearchQueryBuilder

owen-q:
- support searchAfter for nativeSearchQuery

closes https://github.com/spring-projects/spring-data-elasticsearch/issues/2105

### DIFF
--- a/src/test/java/org/springframework/data/elasticsearch/core/paginating/SearchAfterIntegrationTests.java
+++ b/src/test/java/org/springframework/data/elasticsearch/core/paginating/SearchAfterIntegrationTests.java
@@ -15,13 +15,16 @@
  */
 package org.springframework.data.elasticsearch.core.paginating;
 
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import org.elasticsearch.search.sort.SortBuilders;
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/data/elasticsearch/core/paginating/SearchAfterIntegrationTests.java
```diff
@@ -15,13 +15,16 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import org.elasticsearch.search.sort.SortBuilders;
+import org.elasticsearch.search.sort.SortOrder;
```
sothawo, line 27: This pulls in dependencies to Elasticsearch classes in this test, this test is independent of the used query type.

sothawo (changes requested):
We should not use the `SearchAfterIntegrationTests` to test this change. The change is on the `NativeSearchQueryBuilder` and should be tested there. I know there is no test for this builder, but this might be a good opportunity to create it and just test that the `searchAfter` values are set on the `Query` after building it.

owen-q:
> We should not use the `SearchAfterIntegrationTests` to test this change. The change is on the `NativeSearchQueryBuilder` and should be tested there. I know there is no test for this builder, but this might be a good opportunity to create it and just test that the `searchAfter` values are set on the `Query` after building it.

@sothawo 
reflect your PR review.
please review again on changes, thank you

sothawo (approved):

Issue #2062, opened 2022-01-16
Enhance @Query annotated method and methods for namedQueries to have additional info


sothawo:
Coming from this [Stack Overflow question](https://stackoverflow.com/questions/70726155/spring-data-elasticsearch-named-queries-for-agreggation):

Currently the value of a `@Query` annotation or a namedQuery property is sent to Elasticsearch with a [wrapper query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wrapper-query.html) which only accepts the query definition.

But a search request can have addition info like aggregations (_aggs_). 

It might be possible to define these as well in the annotation or the properties file, so that the string in these contains more than just the query part.

Problems to solve:
* This must not break the existing functionality
* We need a different way than a wrapper query to send this
*  Spring Data Elasticsearch must still have the possibility to add fields to the data like offset and size

This needs some more investigation.

nathan-t-lr:
@sothawo hello I am trying to figure out how to submit a pr for this issue, but I don't have access to the repo to write. Am I supposed to fork it or do I need to be granted access as a contributor?

PR #2143, opened 2022-04-21
Feature/query annotation source filter

nathan-t-lr:
This is the implementation of adding a source filter (`includes` and/or `excludes`) to the `Query` annotation. This will allow users to filter fields on an ad-hoc basis only using the additional `includes` and `excludes` arguments in `Query`. A similar approach can be used for solving #2062 to add aggregations.

Relates to #2062

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
@@ -27,6 +27,7 @@ import java.util.Collection;
  * @author Mohsin Husen
  * @author Peter-Josef Meisch
  * @author Steven Pearce
+ * @author Alexander Torres
  */
 
 @Retention(RetentionPolicy.RUNTIME)
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
@@ -17,6 +17,8 @@ package org.springframework.data.elasticsearch.annotations;
 
 import org.springframework.data.annotation.QueryAnnotation;
 import java.lang.annotation.*;
+import java.util.ArrayList;
+import java.util.Collection;
 
 /**
  * Query
@@ -44,6 +46,9 @@ public @interface Query {
 	 * @deprecated since 4.2, not implemented and used anywhere
 	 */
 	String name() default "";
+
+	String includes() default "";
+	String excludes() default "";
 
 	/**
 	 * Returns whether the query defined should be executed as count projection.
--- a/src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchQueryMethod.java
+++ b/src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchQueryMethod.java
@@ -50,6 +50,7 @@ import org.springframework.util.ClassUtils;
  * @author Mark Paluch
  * @author Christoph Strobl
  * @author Peter-Josef Meisch
+ * @author Alexander Torres
  */
 public class ElasticsearchQueryMethod extends QueryMethod {
 
@@ -121,6 +122,22 @@ public class ElasticsearchQueryMethod extends QueryMethod {
 		return new HighlightQuery(
 				org.springframework.data.elasticsearch.core.query.highlight.Highlight.of(highlightAnnotation),
 				getDomainClass());
+	}
+
+	public boolean hasIncludes() {
+		return hasAnnotatedQuery() && !queryAnnotation.includes().equals("");
+	}
+
+	public boolean hasExcludes() {
+		return hasAnnotatedQuery() && !queryAnnotation.excludes().equals("");
+	}
+
+	public String getIncludes() {
+		return queryAnnotation.includes();
+	}
+
+	public String getExcludes() {
+		return queryAnnotation.excludes();
 	}
 
 	/**
--- a/src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQuery.java
+++ b/src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQuery.java
@@ -21,6 +21,8 @@ import org.springframework.data.elasticsearch.core.ElasticsearchOperations;
 import org.springframework.data.elasticsearch.core.SearchHitSupport;
 import org.springframework.data.elasticsearch.core.SearchHits;
 import org.springframework.data.elasticsearch.core.mapping.IndexCoordinates;
+import org.springframework.data.elasticsearch.core.query.FetchSourceFilterBuilder;
+import org.springframework.data.elasticsearch.core.query.SourceFilter;
 import org.springframework.data.elasticsearch.core.query.StringQuery;
 import org.springframework.data.elasticsearch.repository.support.StringQueryUtil;
 import org.springframework.data.repository.query.ParametersParameterAccessor;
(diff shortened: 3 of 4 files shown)


src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQuery.java
```diff
@@ -103,4 +120,23 @@ protected StringQuery createQuery(ParametersParameterAccessor parameterAccessor)
...
+		final String sourceFilterDelim = ",";
+		if (includes.length() != 0) {
+			String includeFilter = stringQueryUtil.replacePlaceholders(includes, parameterAccessor);
+			includeList = includeFilter.split(sourceFilterDelim);
+		}
+		if (excludes.length() != 0) {
+			String excludeFilter = stringQueryUtil.replacePlaceholders(excludes, parameterAccessor);
+			excludeList = excludeFilter.split(sourceFilterDelim);
+		}
+
```
sothawo, line 138: This splitting is not necessary if the values are already coming as a `String[]` from the annotation.

src/test/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQueryUnitTests.java
```diff
@@ -138,6 +141,27 @@ private org.springframework.data.elasticsearch.core.query.Query createQuery(Stri
...
+				"{\"match_all\": {}}"
+		));
+		ParametersParameterAccessor parameterAccessor = mock(ParametersParameterAccessor.class);
+		when(parameterAccessor.getBindableValue(0)).thenReturn("foo");
+		when(parameterAccessor.getBindableValue(1)).thenReturn("bar");
+		SourceFilter resp = elasticsearchStringQuery.createSourceFilter(parameterAccessor);
+		assertThat(resp.getIncludes()).isEqualTo(new String[]{"foo"});
+		assertThat(resp.getExcludes()).isEqualTo(new String[]{"bar"});
+	}
+
```
sothawo, line 164: This is testing the creation of the `SourceFilter`.

What is missing is testing if the values are passed from the `@Query` annotation to the `ElasticsearchQueryMethod` and that the created Query in the end really has the filters set.

I think it would make sense for this use case to have an integration test where it is checked that the values configured in the `@Query` annotation in combination with some method parameters do have the desired response from the ES cluster.

sothawo (changes requested):
Thanks a lot for this PR.
Before submitting a PR please create an issue describing the bug to solve or feature to implement.

I made some comments on your changes. I think this would need an integration test - have a look at the existing integration tests, we currently need to set them up to be running with both the old `RestHighLevelClient`  and the new `ElasticsearchClient`. 

And this logic must be implemented in the reactive code as well, as the `@Query` annotation is used there as well and so must support new functionality.

I think this issue is slightly related to #2062, but that one is more complex. Adding the source filter is a kind of low hanging fruit, as these basically are string arrays that are available on the `BaseQuery` class. They can be manually set on the implementations. Aggregations are more difficult, they are currently created on the `Native*Query` implementations using the Elasticsearch libraries available. To have them defined in the context of a `@Query` annotation would mean that Spring Data Elasticsearch would need to somehow parse an aggs JSON string into the corresponding Elasticsearch builders. That is dependent on the used client library and implementing this before we have removed the `RestHighLevelClient`  integration would mean investing effort in code that will be thrown away within the next year.

src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
```diff
@@ -17,6 +17,8 @@
 
 import org.springframework.data.annotation.QueryAnnotation;
 import java.lang.annotation.*;
+import java.util.ArrayList;
+import java.util.Collection;
```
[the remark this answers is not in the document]
nathan-t-lr: Sorry I'll remove unused imports in the final version

src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
```diff
@@ -45,6 +48,9 @@
 	 */
 	String name() default "";
 
+	String includes() default "";
+	String excludes() default "";
+
```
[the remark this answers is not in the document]
nathan-t-lr: I think when I first did some testing with the @Query notification the List<String> wasn't working (hence the unused imports in earlier comment). I will try with string array if i continue with this implementation.

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQuery.java
nathan-t-lr: got it

src/test/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQueryUnitTests.java
nathan-t-lr: yes I agree, I will do the intergration tests soon

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchQueryMethod.java
```diff
@@ -123,6 +124,22 @@ private HighlightQuery createAnnotatedHighlightQuery() {
 				getDomainClass());
 	}
 
+	public boolean hasIncludes() {
```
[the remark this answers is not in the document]
nathan-t-lr: will do this from now on

nathan-t-lr:
> Thanks a lot for this PR. Before submitting a PR please create an issue describing the bug to solve or feature to implement.
> 
> I made some comments on your changes. I think this would need an integration test - have a look at the existing integration tests, we currently need to set them up to be running with both the old `RestHighLevelClient` and the new `ElasticsearchClient`.
> 
> And this logic must be implemented in the reactive code as well, as the `@Query` annotation is used there as well and so must support new functionality.
> 
> I think this issue is slightly related to #2062, but that one is more complex. Adding the source filter is a kind of low hanging fruit, as these basically are string arrays that are available on the `BaseQuery` class. They can be manually set on the implementations. Aggregations are more difficult, they are currently created on the `Native*Query` implementations using the Elasticsearch libraries available. To have them defined in the context of a `@Query` annotation would mean that Spring Data Elasticsearch would need to somehow parse an aggs JSON string into the corresponding Elasticsearch builders. That is dependent on the used client library and implementing this before we have removed the `RestHighLevelClient` integration would mean investing effort in code that will be thrown away within the next year.

thank you for your review @sothawo! I actually started doing the implementation for aggregations so most of what has been done will change. `SourceFilter` is indeed a much simpler problem then aggregations as I found out, but the way I found 
#2062 was through [a thread discussing the source filter problem](https://stackoverflow.com/questions/71278796/how-exclude-fields-from-source-with-spring-data-elasticsearch). Do you still want me to make a new issue for the SourceFilter by itself and reference in this pr?

sothawo:
> Do you still want me to make a new issue for the SourceFilter by itself and reference in this pr?

Yes, make a separate issue for that and in the PR use "Closes #XXX" to reference the issue

sothawo:
> I actually started doing the implementation for aggregations so most of what has been done will change

Please before doing too much work, let's discuss how you want to implement this, so I can see how this fits into the architecture  and the current migration from the old client to the new one. 

We probably even should have a separate issue for the aggregations to have the discussion there.

nathan-t-lr:
> Please before doing too much work, let's discuss how you want to implement this

Yes, I won't do any testing until the architecture has been approved. It is basically as you said in the previous comment, my approach is have an extra parameter in the `@Query` called `valueParams` which is a json structure. Then I will parse the attributes of it (of which I currently have `aggs` and `_source`) and populate the appropriate builders/fields in the `ElasticsearchStringQuery`. If the builders change in different versions then it shouldn't be hard to migrate. Will commit something soon, perhaps in a new pr for a new issue as you suggested.

nathan-t-lr:
Moving this to its own issue #2146

PR #2147, opened 2022-04-25
Feature/query annotation source filter and aggs

### LINKED ISSUE #2146: Add source filter configuration to @Query repository methods
The `Query` annotation is very limited in that it can only search using matchers but not specify additional configuration like source filters or aggregations. 

Because of this, if advanced searching is desired we must interact directly with `ElasticsearchRestTemplate` using `NativeSearchQuery`s or other means to get the results we want.

We should implement an additional parameter in `@Query` and parse these configurations to add them to the search query. This will be much more convenient and require much less setup time for elasticsearch repository searches.
 
Closes #2062

nathan-t-lr:
These are improvements to the `@Query` annotation that allow a user two additional parameters for querying: `_source` and `aggs`. These are specified through a new parameter `valueParams` which takes a json string as input and is parsed downstream.

Solves issue found on [this thread](https://stackoverflow.com/questions/71278796/how-exclude-fields-from-source-with-spring-data-elasticsearch).
Relates to #2062 
Closes #2146

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
@@ -125,7 +125,8 @@ import org.springframework.util.StringUtils;
  * @author Subhobrata Dey
  * @author Farid Faoudi
  * @author Peer Mueller
- * @since 4.0
+ * @author Alexander Torres
+ * @since 4.4
  */
 class RequestFactory {
 
@@ -938,6 +939,10 @@ class RequestFactory {
 
 		SearchRequest request = new SearchRequest(indexNames);
 		SearchSourceBuilder sourceBuilder = new SearchSourceBuilder();
+		// replace source builder if one specified
+		if (query.getSearchSourceBuilder() != null) {
+			sourceBuilder = query.getSearchSourceBuilder();
+		}
 		sourceBuilder.version(true);
 		sourceBuilder.trackScores(query.getTrackScores());
 		if (hasSeqNoPrimaryTermProperty(clazz)) {
@@ -1014,7 +1019,9 @@ class RequestFactory {
 			sourceBuilder.searchAfter(query.getSearchAfter().toArray());
 		}
 
-		query.getRescorerQueries().forEach(rescorer -> sourceBuilder.addRescorer(getQueryRescorerBuilder(rescorer)));
+		for (RescorerQuery rescorer : query.getRescorerQueries()) {
+			sourceBuilder.addRescorer(getQueryRescorerBuilder(rescorer));
+		}
 
 		if (query.getRequestCache() != null) {
 			request.requestCache(query.getRequestCache());
@@ -1085,6 +1092,12 @@ class RequestFactory {
 
 		if (query.getPreference() != null) {
 			searchRequestBuilder.setPreference(query.getPreference());
+		}
+
+		if (query.getSearchSourceBuilder() != null) {
+			SearchSourceBuilder origBuilder = searchRequestBuilder.request().source();
+			SearchSourceBuilder additionalSource = query.getSearchSourceBuilder();
+			additionalSource.aggregations().getAggregatorFactories().forEach(origBuilder::aggregation);
 		}
 
 		prepareSort(query, searchRequestBuilder, getPersistentEntity(clazz));
--- a/src/main/java/org/springframework/data/elasticsearch/core/query/BaseQuery.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/query/BaseQuery.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.elasticsearch.core.RuntimeField;
@@ -43,6 +44,7 @@ import org.springframework.util.Assert;
  * @author Farid Azaza
  * @author Peter-Josef Meisch
  * @author Peer Mueller
+ * @author Alexander Torres
  */
 public class BaseQuery implements Query {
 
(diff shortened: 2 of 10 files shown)


nathan-t-lr:
@sothawo I have moved the thread to here, some integration tests have been done and it seems to be functioning properly. Let me know what you think :)

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
```diff
@@ -125,7 +125,8 @@
  * @author Subhobrata Dey
  * @author Farid Faoudi
  * @author Peer Mueller
- * @since 4.0
+ * @author Alexander Torres
+ * @since 4.4
```
sothawo, line 129: this would be only changed on _new_ classes or methods

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
```diff
@@ -1014,7 +1019,9 @@ private SearchRequest prepareSearchRequest(Query query, @Nullable Class<?> clazz
 			sourceBuilder.searchAfter(query.getSearchAfter().toArray());
 		}
 
-		query.getRescorerQueries().forEach(rescorer -> sourceBuilder.addRescorer(getQueryRescorerBuilder(rescorer)));
+		for (RescorerQuery rescorer : query.getRescorerQueries()) {
+			sourceBuilder.addRescorer(getQueryRescorerBuilder(rescorer));
+		}
```
sothawo, line 1024: please do not refactor code unrelated to the issue that is solved

src/main/java/org/springframework/data/elasticsearch/core/query/BaseQuery.java
```diff
@@ -26,6 +26,7 @@
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import org.elasticsearch.search.builder.SearchSourceBuilder;
```
sothawo, line 29: no Elasticsearch imports in this class, these imports reference deprecated Elasticsearch classes and cannot be used with the new client.

src/main/java/org/springframework/data/elasticsearch/core/query/FetchSourceFilterBuilder.java
```diff
@@ -15,19 +15,31 @@
  */
 package org.springframework.data.elasticsearch.core.query;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
```
sothawo, line 18: needed here?

src/main/java/org/springframework/data/elasticsearch/core/query/Query.java
```diff
@@ -21,6 +21,7 @@
 import java.util.List;
 import java.util.Optional;
 
+import org.elasticsearch.search.builder.SearchSourceBuilder;
```
sothawo, line 24: no Elasticsearch imports here. This class must be agnostic of the used client implementation.

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchQueryMethod.java
```diff
@@ -123,6 +125,23 @@ private HighlightQuery createAnnotatedHighlightQuery() {
...
+	public boolean hasValueParams() {
+		return hasAnnotatedQuery() && !queryAnnotation.valueParams().equals("");
+	}
+
+	/**
+	 * @since 4.4
+	 * @return the additional configuration for querying
+	 * @throws JsonProcessingException if the json or elasticsearch argument is malformed
+	 */
+	public String getValueParams() throws JsonProcessingException {
```
sothawo, line 141: where would the exception been thrown?

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQuery.java
```diff
@@ -15,18 +15,30 @@
  */
 package org.springframework.data.elasticsearch.repository.query;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.elasticsearch.search.aggregations.*;
+import org.elasticsearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
+import org.elasticsearch.search.builder.SearchSourceBuilder;
```
sothawo, line 23: no Elasticsearch classes here. This will not work with the next Elasticsearch client.

sothawo (changes requested):
I made some principal comments, but taking this discussion to #2146

src/main/java/org/springframework/data/elasticsearch/core/RequestFactory.java
nathan-t-lr: this needed to be done because of line 944. If a lambda is used then `sourceBuilder` needs to be `final`. I think I can write a method to accomplish this though, I will make the change

src/main/java/org/springframework/data/elasticsearch/core/query/FetchSourceFilterBuilder.java
nathan-t-lr: for parsing source filters with missing properties from `@Query` annotation 
e.g.
```
{
  "_source": { "excludes": ["field"] } 
}
```

src/main/java/org/springframework/data/elasticsearch/core/query/BaseQuery.java
nathan-t-lr: if it is deprecated then why is it not tagged? Keep in mind this pr is against `v4.3.x`

src/main/java/org/springframework/data/elasticsearch/core/query/Query.java
nathan-t-lr: why can there be no elasticsearch imports if it is in the elasticsearch core package?

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchQueryMethod.java
nathan-t-lr: sorry will remove this, was thrown in previous implementation

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQuery.java
nathan-t-lr: can you point me to the new client branch you're talking about? I need to review it to have a better context of this problem

nathan-t-lr:
Have made [a new pr](https://github.com/spring-projects/spring-data-elasticsearch/pull/2151) with the recommended changes to main @sothawo. Check out when you can, and if it is good I will implement for the reactive library as well.

PR #2169, opened 2022-05-26
Fix Assert judgment bug

panzhenchao:


sothawo:
Thanks for finding and fixing this. Please, next time first create an issue that is solved by the pull request - as mentioned in the documentation and in the comment template when creating a PR.

Issue #2050, opened 2022-01-06
Use ElasticsearchDateConverter  format same time, but get different timezone result


wd0401:
if now date is : 2022-01-06 16:43:09 , but I get two different results and two different time zone, is this a issue? 

I use the follow codes: 
`
 public static void main(String[] args) {
        ElasticsearchDateConverter elasticsearchDateConverter = ElasticsearchDateConverter.of("yyyy-MM-dd HH:mm:ss");
        // 1
        LocalDateTime localDateTime = LocalDateTime.now();
        Date date = new Date();
        // 2
        String localStr = elasticsearchDateConverter.format(localDateTime);
        System.out.println(localStr); //2022-01-06 16:43:09

        // 3
        String dateStr = elasticsearchDateConverter.format(date);
        System.out.println(dateStr);    // 2022-01-06 08:43:09
}
`

sothawo:
You just discovered one of the reasons why `java.util.Date` should not be used for things like that (better not to use it at all, but the classes from the `java.time` package.

In the format for the converter you just use the date and time info without any timezone information, that's what a `LocalDateTime` is for: just a date with a time and no zone information attached.

A `java.util.Date` is a timestamp that contains the (milli)seconds since Unix epoch (that's January 1st 1970) in the **UTC** timezone. So when you created the `new Date()` (I think you are in a UTC+8 timezone, China, Philippines, Western Australia?) in the Date object the current timestamp in UTC time was recorded. On formatting no timezone is used, so you get this result.

wd0401:
@sothawo tks, got it.

wd0401:
Hi @sothawo ,I thought about this, 
 in the method (org.springframework.data.elasticsearch.core.convert.ElasticsearchDateConverter#format(java.util.Date))
```
	public String format(Date date) {

		Assert.notNull(date, "accessor must not be null");

		return dateFormatter.format(Instant.ofEpochMilli(date.getTime()));
	}
```
would we use it like this? i think it can resovle the  problem for timezone 
```
	public String format(Date date) {

		Assert.notNull(date, "accessor must not be null");

		return dateFormatter.format(LocalDateTime.ofInstant(Instant.ofEpochMilli(date.getTime()), ZoneId.systemDefault()));
	}
```

PR #2172, opened 2022-05-29
Feat:support set timezone

zhaoyunxing92:
I have the same problem #2050 ，I have seen the code setting the `UTC` timezone [ElasticsearchDateConverter.java#L92](https://github.com/spring-projects/spring-data-elasticsearch/blob/c826adb152fb1b00b49f9a8b69db8f969b9ba486/src/main/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchDateConverter.java#L92), but I think this should be set with the system (my time zone is unfortunately `UTC+8`), and 'java.util.Date' is used a lot in development.
 I propose to add this attribute support in `@Field(timezone = "GMT+8")`, I can do this in the next pr.jackson does it `@JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss",timezone = "GMT+8")`

Closes #issue-number

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchDateConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/convert/ElasticsearchDateConverter.java
@@ -36,7 +36,7 @@ import org.springframework.util.Assert;
  * Provides Converter instances to convert to and from Dates in the different date and time formats that elasticsearch
  * understands.
  *
- * @author Peter-Josef Meisch
+ * @author Peter-Josef Meisch zhaoyunxing
  * @since 4.0
  */
 final public class ElasticsearchDateConverter {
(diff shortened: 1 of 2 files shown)


sothawo (changes requested):
Please create an issue before submitting a PR describing the problem so this can be discussed. 
I had a short look at the code you provide, this is changing actual behaviour and so is a breaking change.
We might add additional functionality, but not change the existing one.

zhaoyunxing92:
> Please create an issue before submitting a PR describing the problem so this can be discussed. I had a short look at the code you provide, this is changing actual behaviour and so is a breaking change. We might add additional functionality, but not change the existing one.

I created it #2173

zhaoyunxing92:
@sothawo I would like to ask, what is the meaning of the existence of `converters`.is it possible to directly return to`new ElasticsearchDateConverter(forPattern(subPatterns[0].trim()))` 
``` java
public static ElasticsearchDateConverter of(String pattern) {

    Assert.hasText(pattern, "pattern must not be empty");

    String[] subPatterns = pattern.split("\\|\\|");

    return converters.computeIfAbsent(subPatterns[0].trim(), p -> new ElasticsearchDateConverter(forPattern(p)));
}
```

PR #2254, opened 2022-08-06
Implement SourceFilters annotation.

### LINKED ISSUE #1280: Add include and exclude source parameters to Query annotation [DATAES-707]
**[sothawo](https://jira.spring.io/secure/ViewProfile.jspa?name=sothawo)** opened **[DATAES-707](https://jira.spring.io/browse/DATAES-707?redirect=false)** and commented

Adding _source include and exclude parameters to the `@Query` annotation would allow to build queries that have these arguments set

---

**Reference URL:** https://stackoverflow.com/questions/59190295/annotation-query-and-return-only-specific-nested-field-with-spring-data-elastics

### LINKED ISSUE #2062: Enhance @Query annotated method and methods for namedQueries to have additional info

### LINKED ISSUE #2146: Add source filter configuration to @Query repository methods

sothawo:
This now is implemented as a separate annotation `@SourceFilters`, not as an additional parameter to `@Query`.

This makes this available not only to `@Query`  annotated repository methods, but to normal repository methods as well.

Closes #1280
Closes #2062
Closes #2146
Closes #2147
Closes #2151

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Query.java
@@ -15,8 +15,14 @@
  */
 package org.springframework.data.elasticsearch.annotations;
 
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.core.annotation.AliasFor;
 import org.springframework.data.annotation.QueryAnnotation;
-import java.lang.annotation.*;
 
 /**
  * Query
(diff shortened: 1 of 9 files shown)


nathan-t-lr:
Thank you for merging this @sothawo , I have been so busy since the original dev work for this feature and was only just getting the time to look at it again. We will start to use this annotation on my team :)

sothawo:
One more info: I kept the separate annotation instead of adding it to `@Query` so it can be used on "normal" repository functions as well.

nathan-t-lr:
> One more info: I kept the separate annotation instead of adding it to `@Query` so it can be used on "normal" repository functions as well.

Yes I remember that was the plan last time we left off in development, totally fine.

On a separate note, is there some way I can be added as a contributor to the repo? I know my name is tagged as an author but its a very common name.

PR #2288, opened 2022-09-03
Fix NPE in RequestFactory when language in UpdateQuery is not set (4.4.x)

### LINKED ISSUE #2287: Fix NPE in RequestFactory when language in UpdateQuery is not set (4.4.x).

sothawo:
Closes #2287


37176427:
I found the same problem in the **updateByQuery**:
 
requestFactory.updateByQueryRequest(query, index);

PR #2305, opened 2022-09-22
Support partially update document by entity.

### LINKED ISSUE #2304: update document partially by specified entity
Now if we wanna update a document, a UpdateQuery must be constructed manually, but this is a little unnecessary. Especially for the condation that:
1. get a document from index;
2. update some field of the document;
3. write the doc back;

Problems are:
1. if write the doc back using `save(ENTITY)`, values of fields which are unmapped in the entity will be discarded;
2. if write the doc back using `update(UpdateQuery, IndexCoordinates)`, a UpdateQuery and a IndexCoordinates must be constructed manually;

So i think `update(T entity)` is necessary, and spring-data-elasticsearch can construct the UpdateQuery and IndexCoordinates automatically, which is very handy.

puppylpg:

Closes #2304

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
@@ -48,6 +48,7 @@ import org.springframework.data.elasticsearch.core.query.MoreLikeThisQuery;
 import org.springframework.data.elasticsearch.core.query.Query;
 import org.springframework.data.elasticsearch.core.query.SeqNoPrimaryTerm;
 import org.springframework.data.elasticsearch.core.query.UpdateQuery;
+import org.springframework.data.elasticsearch.core.query.UpdateResponse;
 import org.springframework.data.elasticsearch.core.routing.DefaultRoutingResolver;
 import org.springframework.data.elasticsearch.core.routing.RoutingResolver;
 import org.springframework.data.elasticsearch.support.VersionInfo;
@@ -74,6 +75,7 @@ import org.springframework.util.Assert;
  * @author Subhobrata Dey
  * @author Steven Pearce
  * @author Anton Naydenov
+ * @author Haibo Liu
  */
 public abstract class AbstractElasticsearchTemplate implements ElasticsearchOperations, ApplicationContextAware {
 
@@ -305,7 +307,7 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 	@Override
 	public String delete(Object entity, IndexCoordinates index) {
 		String entityId = getEntityId(entity);
-		Assert.notNull(entityId, "entity must have an if that is notnull");
+		Assert.notNull(entityId, "entity must have an id that is notnull");
 		return this.delete(entityId, index);
 	}
 
@@ -468,6 +470,25 @@ public abstract class AbstractElasticsearchTemplate implements ElasticsearchOper
 		return getRequiredPersistentEntity(clazz).getIndexCoordinates();
 	}
 
+	@Override
+	public <T> UpdateResponse update(T entity) {
+		return update(this.buildUpdateQueryByEntity(entity), this.getIndexCoordinatesFor(entity.getClass()));
+	}
+
+	protected <T> UpdateQuery buildUpdateQueryByEntity(T entity) {
+		String id = this.getEntityId(entity);
+		Assert.notNull(entity, "entity must have an id that is notnull");
+
+		UpdateQuery.Builder updateQueryBuilder = UpdateQuery.builder(id)
+				.withDocument(elasticsearchConverter.mapObject(entity));
+
+		String routing = this.getEntityRouting(entity);
+		if (Objects.nonNull(routing)) {
+			updateQueryBuilder.withRouting(routing);
+		}
+		return updateQueryBuilder.build();
+	}
+
 	protected <T> T updateIndexedObject(T entity, IndexedObjectInformation indexedObjectInformation) {
 
 		ElasticsearchPersistentEntity<?> persistentEntity = elasticsearchConverter.getMappingContext()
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchIntegrationTests.java
```diff
@@ -177,6 +178,20 @@ protected abstract Query getMatchAllQueryWithIncludesAndInlineExpressionScript(@
 
 	protected abstract Query getQueryWithRescorer();
 
+	@Test
```
sothawo, line 181:
```suggestion
	@Test // #2304
```

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchIntegrationTests.java
```diff
@@ -1499,6 +1514,33 @@ public void shouldDeleteIndexForGivenEntity() {
...
+		// modify the entity
+		sampleEntity.setMessage(messageAfterUpdate);
+
+		// when
+		operations.update(sampleEntity);
+
+		// then
+		SampleEntity indexedEntity = operations.get(documentId, SampleEntity.class,
+				IndexCoordinates.of(indexNameProvider.indexName()));
+		assertThat(indexedEntity.getMessage()).isEqualTo(messageAfterUpdate);
```
sothawo, line 1541:
```suggestion
		// given
		String documentId = nextIdAsString();
		String messageBeforeUpdate = "some test message";
		String messageAfterUpdate = "test message";

		SampleEntity sampleEntity = SampleEntity.builder().id(documentId).message(messageBeforeUpdate).type("some type")
				.version(System.currentTimeMillis()).build();
		operations.save(sampleEntity);

		// modify the entity
		sampleEntity.setMessage(messageAfterUpdate);
		sampleEntity.setType(null);

		// when
		operations.update(sampleEntity);

		// then
		SampleEntity indexedEntity = operations.get(documentId, SampleEntity.class);
		assertThat(indexedEntity.getType()).isEqualTo("some type");
		assertThat(indexedEntity.getMessage()).isEqualTo(messageAfterUpdate);
```

Just modifying one property is not testing the partial update, you'd get the same result when doing a normal `save`. You need do actually set some property to `null` to make the update partial and check, thta this null property was not updated.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -508,7 +529,7 @@ ElasticsearchPersistentEntity<?> getRequiredPersistentEntity(Class<?> clazz) {
 	}
 
 	@Nullable
-	private String getEntityId(Object entity) {
+	public String getEntityId(Object entity) {
```
sothawo, line 533: That change is not necessary for this issue.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -468,6 +470,25 @@ public IndexCoordinates getIndexCoordinatesFor(Class<?> clazz) {
 		return getRequiredPersistentEntity(clazz).getIndexCoordinates();
 	}
 
+	@Override
+	public <T> UpdateResponse update(T entity) {
+		return update(this.buildUpdateQueryByEntity(entity), this.getIndexCoordinatesFor(entity.getClass()));
```
sothawo, line 475:
```suggestion
		return update(buildUpdateQueryByEntity(entity), getIndexCoordinatesFor(entity.getClass()));
```
we don't qualify method calls with `this.`.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -468,6 +470,25 @@ public IndexCoordinates getIndexCoordinatesFor(Class<?> clazz) {
...
+	}
+
+	protected <T> UpdateQuery buildUpdateQueryByEntity(T entity) {
+		String id = this.getEntityId(entity);
+		Assert.notNull(entity, "entity must have an id that is notnull");
```
sothawo, line 480:
```suggestion
		Assert.notNull(entity, "entity must not be null");
		
		String id = this.getEntityId(entity);
		Assert.notNull(id, "entity must have an id that is not null");
```

assert the entity and then the id

sothawo (approved):
Thanks for this PR. I have commented some minor things, I will fix them with a Polishing commit after merging.

sothawo:
Sorry, had to revert the merge. I missed that it was targeting the wrong branch. 
We do not add new features in published versions, only on the main branch, so this should be merged into the `main` branch.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
puppylpg: It's true this modifycation is unrelated with this PR. However, 

before this PR, I update an entity using spring-data-elasticsearch in this way:
- to get `_id`: ElasticsearchRestTemplate#getEntityId
- to get `_routing`: ElasticsearchRestTemplate#getEntityRouting
- to get `_source`: elasticsearchConverter.mapObject(entity)

(Then combine id/routing/source to a UpdateQuery.)

The latter two methods have public access while the first one is `private` thus **must be copied from `ElasticsearchRestTemplate` to my class** manually to use. If it was `public`, this copy won't happen.

If this PR is merged, seems that there is no need to access `getEntityId` in outer class again. However, `getRouting(entity)` is similar to this method in the same class and be marked as `public`, which is convenient to use. I think it may be better to be accessible for this method too.

Anyway, I'll not modify this line to `public` in new PR, but leave it for your consideration.

puppylpg (commented):
changes are accepted

PR #2436, opened 2023-01-25
Overload multiSearch method to support list of IndexCoordinates

### LINKED ISSUE #2434: Add support for separate IndexCoordinates query in multiSearch.
The API 

`multiSearch(List<? extends Query> queries, List<Class<?>> classes,
			IndexCoordinates index)`

takes an IndexCoordinates and sends in with each query in the multiSearch. ElasticSearch supports different indexes per query in the multisearch so we should be able to send in a list of IndexCoordinates that is iterated along with the query and class.

AIRahimi:
Pretty much copy/paste with the exception of one more list (IndexCoordinates) that is iterated along with the classes and added per query. 

Could've (maybe should've) refactored existing methods to avoid duplicate fragments, but did not want to change existing code too much since this is my first contribution.

Closes #2434


sothawo:
Thanks a lot for this contribution!

PR #2505, opened 2023-03-27
Allow IndexCoordinates in repositories

ls-urs-keller:
It is common to have multiple indexes with the same document structure. E.g. rolling indices over time and having a wildcard alias. Sometimes one wants to select different IndexCoordinates than specified in `@Document(indexName = "...", ...)`. Also you might want to use the same Repository to do cross cluster searches and for this need to specify the IndexCoordinates.

It would therefore be helpful to support IndexCoordinates as an argument to repository methods the same way ScrollPosition, Sort and Pageable are handled.

This PR is an attempt to introduce this change. I keep it as draft to get some comments I can add tests to it if we can agree on the change being valid.

- [ ] **There is a ticket in the bug tracker for the project in our [issue tracker](https://github.

Closes #issue-number


src/main/java/org/springframework/data/elasticsearch/repository/query/AbstractElasticsearchRepositoryQuery.java
```diff
@@ -32,8 +30,6 @@
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 
-import java.util.Collections;
-
 /**
  * AbstractElasticsearchRepositoryQuery
  *
```
sothawo, line 35: add yourself to the authors in all the files you change

src/main/java/org/springframework/data/elasticsearch/repository/query/AbstractElasticsearchRepositoryQuery.java
```diff
@@ -74,15 +70,17 @@ public QueryMethod getQueryMethod() {
...
 		Query query = createQuery(parameters);
 
-		IndexCoordinates index = elasticsearchOperations.getIndexCoordinatesFor(clazz);
+		IndexCoordinates index = parameterAccessor
+				.getIndexCoordinatesOrDefaults(elasticsearchOperations.getIndexCoordinatesFor(clazz));
 
 		Object result = null;
 
 		if (isDeleteQuery()) {
+			index = elasticsearchOperations.getIndexCoordinatesFor(clazz);
```
sothawo, line 83: don't need this line, index is already set above

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchParameterAccessor.java
```diff
@@ -26,7 +28,9 @@ public interface ElasticsearchParameterAccessor extends ParameterAccessor {
 	/**
 	 * Returns the raw parameter values of the underlying query method.
 	 *
-	 * @return
+	 * @return values.
 	 */
 	Object[] getValues();
+
+	IndexCoordinates getIndexCoordinatesOrDefaults(@NonNull IndexCoordinates defaults);
```
sothawo, line 35: add javadoc to the new method

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchParameters.java
```diff
@@ -16,54 +16,89 @@
...
+
 	/**
 	 * Custom {@link Parameter} implementation adding parameters of type {@link Distance} to the special ones.
 	 *
 	 * @author Christoph Strobl
 	 */
-	class ElasticsearchParameter extends Parameter {
+	static class ElasticsearchParameter extends Parameter {
+
+		private final boolean indexCoordinatesParameter;
```
sothawo, line 83:
```suggestion
		private final boolean isIndexCoordinatesParameter;
```

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchParameters.java
```diff
@@ -16,54 +16,89 @@
...
+		for (ElasticsearchParameter parameter : this) {
+			if (parameter.isIndexCoordinatesParameter()) {
+				foundIndices.add(index);
+			}
+			index++;
+		}
+		if (foundIndices.size() > 1) {
+			throw new IllegalArgumentException(this + " can only contain at most one IndexCoordinates parameter.");
+		}
+		return foundIndices.isEmpty() ? -1 : foundIndices.get(0);
```
sothawo, line 60: alternate way without a list

```
		int indexCoordinatesIndex = -1;
		int index = 0;
		for (ElasticsearchParameter parameter : this) {
			if (parameter.isIndexCoordinatesParameter()) {
				if(indexCoordinatesIndex != -1) {
					throw new IllegalArgumentException(this + " can only contain at most one IndexCoordinates parameter.");
				} else {
					indexCoordinatesIndex = index;
				}
			}
			index++;
		}
		return indexCoordinatesIndex;
```

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchParametersParameterAccessor.java
```diff
@@ -36,13 +38,25 @@ class ElasticsearchParametersParameterAccessor extends ParametersParameterAccess
...
+		return (ElasticsearchParameters) super.getParameters();
+	}
+
 	@Override
 	public Object[] getValues() {
 		return values.toArray();
 	}
+
+	@Override
+	public IndexCoordinates getIndexCoordinatesOrDefaults(@NonNull IndexCoordinates defaults) {
```
sothawo, line 56:
```suggestion
	public IndexCoordinates getIndexCoordinates(@NonNull IndexCoordinates defaults) {
```

I would not put that in the method name in addition to the parameter name

sothawo (commented):
Looks good to me, I left some comments, but only minor things.

Needs integration tests, but as you wrote, still a draft.

ls-urs-keller:
> Hi, any progress with this one? I could help with integration tests if needed :)

@Bartolg sitting in my inbox, contribution of tests would be welcome

Issue #2542, opened 2023-05-10
"NativeQueryBuilder" class : the "withSearchExtensions" function does not set the expected field


pierre-mazieres-semarchy:
On last 5.0.5 version,
In the "org.springframework.data.elasticsearch.client.elc" package,
There is the "NativeQueryBuilder" class.

This class has a field named "searchExtensions".
`private Map<String, JsonData> searchExtensions = new LinkedHashMap<>()`

There is a "withSearchExtensions" setter function for this field.
`public NativeQueryBuilder withSearchExtensions(Map<String, JsonData> searchExtensions)`

In order for the "searchExtensions" field to be set,
we have to correct this code line `searchExtensions.putAll(searchExtensions)`
to `this.searchExtensions.putAll(searchExtensions)`

For information, I created a pull-request
[#2543](https://github.com/spring-projects/spring-data-elasticsearch/pull/2543)

PR #2543, opened 2023-05-10
"NativeQueryBuilder" class : the "withSearchExtensions" function does not set the expected field

pierre-mazieres-semarchy:
This a proposal to correct this ticket : [#2542](https://github.com/spring-projects/spring-data-elasticsearch/issues/2542)


sothawo:
Thanks for finding and fixing this.

PR #2596, opened 2023-06-19
Fix some typos in object mapping doc.

deejay1:


sothawo:
Thanks for finding and fixing this

PR #2705, opened 2023-09-21
 Backport add IndicesOptions to search request.

### LINKED ISSUE #2639: No way to ignore unavailable indices after upgrade from 4.x
Since the new API doesn't support the old indices options, I have no practical way to tell the server to ignore a missing index when doing a search in multiple indices. So when I do a search with an index missing, I get an error instead of no results for that particular index (a totally acceptable service degradation).

However `co.elastic.clients.elasticsearch.core.SearchRequest.Builder` _does_ allow setting `ignoreUnavailable` and it works for both older (7.17.x) and newer (8.8.x) ES servers I tested it on when setting the option via debugger.

Is there a way to expose this option somewhere if not in `BaseQueryBuilder` then at least `NativeQueryBuilder` I use in conjunction with `ElasticsearchTemplate`? Or some workaround other than preemptively fetching and caching the available indices.

### LINKED ISSUE #2641: Apply IndicesOption settings to Elasticsearch query.
When the initial integration of the new Elasticsearch client was implemented, the search request builder did not yet support properties of the indices options and wildcard states like the old one did. Therefore the property `org.springframework.data.elasticsearch.core.query.Query#getIndicesOptions` was ignored when building a request.

In the meantime most of the previously available options can - and should - be set.

paulrizzo:
Closes #2641
 Closes #2639

@sothawo
We are working to upgrade our clients on the 4.4.x line before making the upgrade to the 5.x along with ES server 8.x and encountered the issue as described in #2639 which you have graciously already addressed in the latest version. I'm just looking to backport this change to help others in the same boat upgrading to the 4.4.x line to utilize the new client before making the jump forward to the next major version.

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
@@ -19,6 +19,7 @@ import static org.springframework.data.elasticsearch.client.elc.TypeUtils.*;
 import static org.springframework.util.CollectionUtils.*;
 
 import co.elastic.clients.elasticsearch._types.Conflicts;
+import co.elastic.clients.elasticsearch._types.ExpandWildcard;
 import co.elastic.clients.elasticsearch._types.FieldValue;
 import co.elastic.clients.elasticsearch._types.InlineScript;
 import co.elastic.clients.elasticsearch._types.OpType;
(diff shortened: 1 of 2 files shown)


sothawo:
Enhancements and especially API changes and additions are not backported in Spring Data Elasticsearch. For released branches only bugfixes are backported. 

Besides that, branch 4.4 any is anyway out of maintenance since May 2023; the only updates on this branch are minor version updates of the Elasticsearch dependency because Spring Boot 2.7 (Spring Boot has a longer maintenance period than Spring Data Elasticsearch) pulls this in and we adjust to pull in the same dependencies as they do, besides that branch 4.4 will not get any further changes.

PR #2793, opened 2023-12-08
support highlight_query

### LINKED ISSUE #2636: Add support for highlight queries in highlighting
### Problem
I want to replace ElasticSearchRestTemplate for ElasticSearchTemplate in order to get rid of deprecated RestHighLevelClient.

There's a use case where a highlight query is needed.
Elasticsearch allows to specify a highlight query as explained in [Elasticseach documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html#specify-highlight-query
)
### Version
Spring Data Elasticsearch 4.4.5

### Issue
`NativeSearchQueryBuilder` allowed using it like this:
`new NativeSearchQueryBuilder().withHighlightBuilder(new HighlightBuilder().highlightQuery(_here_comes_the_query_));`

But `NativeQueryBuilder` does not allow it, as `withHighlight` method from `BaseQueryBuilder` accepts a `Highlight` and a class instead of a `HighlightBuilder`

Is there something I didn't see or is the possibility to add a highlight query really missing?

Please note Elasticsearch provides a `Hightlight.Builder` class that could be used as a replacement for `HighlightBuilder`

puppylpg:

Closes #2636

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
@@ -1647,7 +1647,7 @@ class RequestConverter {
 	}
 
 	@Nullable
-	public co.elastic.clients.elasticsearch._types.query_dsl.Query getQuery(@Nullable Query query,
+	co.elastic.clients.elasticsearch._types.query_dsl.Query getQuery(@Nullable Query query,
 			@Nullable Class<?> clazz) {
 
 		if (query == null) {
--- a/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
@@ -105,6 +105,7 @@ import org.springframework.util.StringUtils;
  * @author Sascha Woo
  * @author cdalxndr
  * @author scoobyzhang
+ * @author Haibo Liu
  * @since 4.4
  */
 @SuppressWarnings("ClassCanBeRecord")
@@ -1494,7 +1495,7 @@ class RequestConverter {
 	private void addHighlight(Query query, SearchRequest.Builder builder) {
 
 		Highlight highlight = query.getHighlightQuery()
-				.map(highlightQuery -> new HighlightQueryBuilder(elasticsearchConverter.getMappingContext())
+				.map(highlightQuery -> new HighlightQueryBuilder(elasticsearchConverter.getMappingContext(), this)
 						.getHighlight(highlightQuery.getHighlight(), highlightQuery.getType()))
 				.orElse(null);
 
@@ -1504,7 +1505,7 @@ class RequestConverter {
 	private void addHighlight(Query query, MultisearchBody.Builder builder) {
 
 		Highlight highlight = query.getHighlightQuery()
-				.map(highlightQuery -> new HighlightQueryBuilder(elasticsearchConverter.getMappingContext())
+				.map(highlightQuery -> new HighlightQueryBuilder(elasticsearchConverter.getMappingContext(), this)
 						.getHighlight(highlightQuery.getHighlight(), highlightQuery.getType()))
 				.orElse(null);
 
@@ -1646,7 +1647,7 @@ class RequestConverter {
 	}
 
 	@Nullable
-	private co.elastic.clients.elasticsearch._types.query_dsl.Query getQuery(@Nullable Query query,
+	public co.elastic.clients.elasticsearch._types.query_dsl.Query getQuery(@Nullable Query query,
 			@Nullable Class<?> clazz) {
 
 		if (query == null) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/data/elasticsearch/core/query/highlight/HighlightCommonParameters.java
```diff
@@ -15,10 +15,12 @@
  */
 package org.springframework.data.elasticsearch.core.query.highlight;
 
+import co.elastic.clients.elasticsearch._types.query_dsl.Query;
```
sothawo, line 18: This would pull Elasticsearch classes into the code outside of the `org.springframework.data.elasticsearch.client.elc` package. We cannot do that. There might be different clients in the future - as there have been in the past, and there are projects depending on Spring Data Elasticsearch (like spring-data-opensearch), they would be forced to use classes from the Elasticsearch client library.

puppylpg: So there should be a highlight query wrapper in package `org.springframework.data.elasticsearch.client.elc` like `NativeQuery`, and `HighlightCommonParameters` just depends on this wrapper right?

puppylpg: Seems that I misunderstood what you mean before. Things are not that simple and using a wrapper still causes the `HighlightCommonParameters` in spring-data-elsticsearch `core` package depending on classes in elasticsearch-java indirectly.

Maybe:
1. there should be a brand new class (say `A`) for highlight query used by `HighlightCommonParameters` in `core` which has nothing to do with the classes in elasticsearch-java client;
2. a corresponding converter(say `A_Processor`) in `org.springframework.data.elasticsearch.client.elc` will handle the conversion from `A` to `co.elastic.clients.elasticsearch._types.query_dsl.Query` which is used by highlight query class in elasticsearch-java;

puppylpg: I think the above `A` should be `org.springframework.data.elasticsearch.core.query.Query`, and there is already a way to convert from  `org.springframework.data.elasticsearch.core.query.Query` to  ` `org.springframework.data.elasticsearch.core.query.Query` in `RequestConverter`.

puppylpg: I think the above `A` should be `org.springframework.data.elasticsearch.core.query.Query`, and there is already a way to convert from  `org.springframework.data.elasticsearch.core.query.Query` to  ` `org.springframework.data.elasticsearch.core.query.Query` in `RequestConverter`.

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchIntegrationTests.java
```diff
@@ -2964,6 +2966,105 @@ void shouldReturnHighlightFieldsInSearchHit() {
...
+		Query query = getBuilderWithTermQuery("message", "message") //
+				.withHighlightQuery(
+						new HighlightQuery(new Highlight(singletonList(new HighlightField("message", sameHighlightQuery))), HighlightEntity.class)
+				)
+				.build();
+		SearchHits<HighlightEntity> searchHits = operations.search(query, HighlightEntity.class, index);
+
+		assertThat(searchHits).isNotNull();
+		assertThat(searchHits.getSearchHits()).hasSize(1);
+
```
sothawo, line 2993: Please split the different parts of this test into separate test methods

src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
```diff
@@ -1646,7 +1647,7 @@ private void prepareNativeSearch(NativeQuery query, MultisearchBody.Builder buil
 	}
 
 	@Nullable
-	private co.elastic.clients.elasticsearch._types.query_dsl.Query getQuery(@Nullable Query query,
+	public co.elastic.clients.elasticsearch._types.query_dsl.Query getQuery(@Nullable Query query,
```
sothawo, line 1650:
```suggestion
	co.elastic.clients.elasticsearch._types.query_dsl.Query getQuery(@Nullable Query query,
```
changing the visibility to package scope is enough here, don't expose it to public

sothawo (changes requested):
Looks great, two minor remarks

src/main/java/org/springframework/data/elasticsearch/core/query/highlight/HighlightCommonParameters.java
sothawo: You found out by yourself what to change, ne need for me to write that 😁

sothawo (approved):

PR #2805, opened 2023-12-21
ClientConfigurer is only available in MaybeSecureClientConfigurationBuilder

dadoonet:
The documentation code does not compile and I'm not sure if it's a bug in the code or a miss in the documentation.

When you want to configure a client, the doc says to use:

```java
ClientConfiguration.builder().withClientConfigurer( // ...
```

But `withClientConfigurer(ClientConfiguration.ClientConfigurationCallback<?> clientConfigurer)` is only available in `TerminalClientConfigurationBuilder` interface.

And `ClientConfiguration.builder()` returns a `ClientConfigurationBuilderWithRequiredEndpoint` interface.

- [ ] **There is a ticket in the bug tracker for the project in our [issue tracker](https://github.com/spring-projects/spring-data-elasticsearch/issues)**. Add the issue number to the _Closes #issue-number_ line below

Closes #issue-number

### DIFF
--- a/src/main/antora/modules/ROOT/pages/elasticsearch/clients.adoc
+++ b/src/main/antora/modules/ROOT/pages/elasticsearch/clients.adoc
@@ -192,6 +192,7 @@ This callback provides a `org.elasticsearch.client.RestClientBuilder` that can b
 [source,java]
 ----
 ClientConfiguration.builder()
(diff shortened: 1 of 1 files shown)


sothawo:
You're right, that's a documentation error. Having a configuration without specifying where to connect to makes no sense, so that's the first needed step. Thanks for finding and adding this.

PR #2806, opened 2023-12-21
Expose search shard statistics in search hits

### LINKED ISSUE #2605: Expose the method to get total shards & successful shards from SearchHits response
Hi Team,

We have one use case where we have to check and need to do some handling on the basis of total shards and successful shards. So it will be good if we can expose them from org.springframework.data.elasticsearch.core.SearchHits

`/*
public interface SearchHits<T> extends Streamable<SearchHit<T>> {

	/**
	 * @return the aggregations.
	 */
	@Nullable
	AggregationsContainer<?> getAggregations();

	/**
	 * @return the maximum score
	 */
	float getMaxScore();

	/**
	 * @param index position in List.
	 * @return the {@link SearchHit} at position {index}
	 * @throws IndexOutOfBoundsException on invalid index
	 */
	SearchHit<T> getSearchHit(int index);

	/**
	 * @return the contained {@link SearchHit}s.
	 */
	List<SearchHit<T>> getSearchHits();

	/**
	 * @return the number of total hits.
	 */
	long getTotalHits();

	/**
	 * @return the relation for the total hits
	 */
	TotalHitsRelation getTotalHitsRelation();

	/**
	 * @return true if aggregations are available
	 */
	default boolean hasAggregations() {
		return getAggregations() != null;
	}

	/**
	 * @return whether the {@link SearchHits} has search hits.
	 */
	default boolean hasSearchHits() {
		return !getSearchHits().isEmpty();
	}

	/**
	 * @return the suggest response
	 * @since 4.3
	 */
	@Nullable
	Suggest getSuggest();

	/**
	 * @return wether the {@link SearchHits} has a suggest response.
	 * @since 4.3
	 */
	default boolean hasSuggest() {
		return getSuggest() != null;
	}

	/**
	 * @return an iterator for {@link SearchHit}
	 */
	default Iterator<SearchHit<T>> iterator() {
		return getSearchHits().iterator();
	}

}
`

puppylpg:

Closes #2605

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/core/SearchShardStatistics.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/SearchShardStatistics.java
@@ -20,37 +20,5 @@ import org.springframework.lang.Nullable;
 /**
  * @author Haibo Liu
  */
-public class SearchShardStatistics {
-	private final Number failed;
-
-	private final Number successful;
-
-	private final Number total;
-
-	@Nullable
-	private final Number skipped;
-
-	public SearchShardStatistics(Number failed, Number successful, Number total, @Nullable Number skipped) {
-		this.failed = failed;
-		this.successful = successful;
-		this.total = total;
-		this.skipped = skipped;
-	}
-
-	public Number getFailed() {
-		return failed;
-	}
-
-	public Number getSuccessful() {
-		return successful;
-	}
-
-	public Number getTotal() {
-		return total;
-	}
-
-	@Nullable
-	public Number getSkipped() {
-		return skipped;
-	}
+public record SearchShardStatistics(Number failed, Number successful, Number total, @Nullable Number skipped) {
 }
--- a/src/main/java/org/springframework/data/elasticsearch/core/SearchShardStatistics.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/SearchShardStatistics.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2013-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * @author Haibo Liu
+ */
+public class SearchShardStatistics {
+	private final Number failed;
+
+	private final Number successful;
+
+	private final Number total;
+
+	@Nullable
+	private final Number skipped;
+
+	public SearchShardStatistics(Number failed, Number successful, Number total, @Nullable Number skipped) {
+		this.failed = failed;
+		this.successful = successful;
+		this.total = total;
+		this.skipped = skipped;
+	}
+
+	public Number getFailed() {
+		return failed;
+	}
+
+	public Number getSuccessful() {
+		return successful;
+	}
+
+	public Number getTotal() {
+		return total;
+	}
+
+	@Nullable
+	public Number getSkipped() {
+		return skipped;
+	}
+}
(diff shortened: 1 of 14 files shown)


src/main/java/org/springframework/data/elasticsearch/core/SearchShardStatistics.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.core;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * @author Haibo Liu
+ */
+public record SearchShardStatistics(Number failed, Number successful, Number total, @Nullable Number skipped) {
```
sothawo, line 23: wouldn't it make sense to add the list of ShardFailures as well?

src/main/java/org/springframework/data/elasticsearch/repository/query/AbstractElasticsearchRepositoryQuery.java
```diff
@@ -110,13 +107,7 @@ public Object execute(Object[] parameters) {
...
 			if (parameterAccessor.getPageable().isUnpaged()) {
 				int itemCount = (int) elasticsearchOperations.count(query, clazz, index);
-
-				if (itemCount == 0) {
-					result = new SearchHitsImpl<>(0, TotalHitsRelation.EQUAL_TO, Float.NaN, null,
-							query.getPointInTime() != null ? query.getPointInTime().id() : null, Collections.emptyList(), null, null);
-				} else {
-					query.setPageable(PageRequest.of(0, Math.max(1, itemCount)));
-				}
+				query.setPageable(PageRequest.of(0, Math.max(1, itemCount)));
```
sothawo, line 110: in line 115 result now is always null, so the if condition can be removed

src/test/java/org/springframework/data/elasticsearch/client/elc/SearchDocumentResponseBuilderUnitTests.java
```diff
@@ -108,4 +112,32 @@ void shouldGetPhraseSuggestion() throws JSONException {
...
+				.hits(new ArrayList<>())
+				.build();
+
+		ShardStatistics shards = new ShardStatistics.Builder()
+				.total(15)
+				.successful(15)
+				.skipped(0)
+				.failed(0)
+				.build();
+
```
sothawo, line 132: add failure objects

sothawo (changes requested):
Thank you for implementing this. Please add the failure objects to the new class, otherwise just some minor comments.

src/main/java/org/springframework/data/elasticsearch/core/SearchShardStatistics.java
puppylpg: I found this a little awkward: if `ShardFailure` is added into this class, shall we rewrite `ShardFailure`(and `ErrorCause`, `JsonData map` inside it) of our own? or shall we just expose these classes written by `elasticsearch-java` to our users?

I think spring-data-elasticsearch obeys the first rule and almost all the classes exposed to the users are written by spring-data-elasticsearch itself. However, I also notice that when querying aggregations, the class `co.elastic.clients.elasticsearch._types.aggregations.Aggregate` in `elasticsearch-java` is exposed to the users directly.

So which implementation should we choose?

sothawo: It's awkward, yes. And there are already places where Elasticsearch failures are mapped to corresponding classes from Spring Data Elasticsearch. The ugly thing is, that the Elasticsearch client library for this parts  is autogenerated from a schema definition, but this generation is not able to reuse classes, so they redefine them multiple times.

As for exposing Elasticsearch classes: No, we cannot do that, because not every user is using Elasticsearch. Spring Data Elasticsearch also is the base project for [spring-data-opensearch](https://github.com/opensearch-project/spring-data-opensearch) where basically the low level implementation is changed, but the high level stuff is used from Spring Data Elasticsearch. And we cannot introduce Elasticsearch classes into that code.

In version 4, we had Elasticsearch classes happily over the codebase, but with the refactoring needed for the new client in 5 all the code dependent on the old classes went into the `.erhlc.` package and the new one to `.elc.`. There first was the idea of eventually adding a `.orhlc.` for the opensearch rest high level client, but after a call with people from AWS it was decided that all the client specific code has to go into a separate package, they provide the opensearch part in their own project. And everything outside of this pacakge must be independent of a client library.

As for the aggregations: In the Spring Data Elasticsearch code, only interfaces are returned, the concrete implementations come from the client implementations. The main reason why there is no client independent mapping for aggregations is the shere amount of classes that would be needed to map, the last time I checked there were almost 190 classes in the Elasticsearch package. And that needs to be done in every client implementation.

puppylpg: Thanks for the detailed explanation. I've seen the corresponding parts, and luckily at least the error cause is reused in the elasticsearch client schema definition :smiling_face_with_tear:

PR #2826, opened 2024-01-13
Add support for SpEL in `@Query`

### LINKED ISSUE #2083: Add support for SpEl in @Query
It would be cool if elasticsearch querys can be used @Query using Spring Expression Language, similar like spring-data-jpa uses on its @Query annotated methods, this to be able to reduce the amount of params when the conditional keywords are scalating quickly.

see [SpEL with @Query in spring data elasticsearch in StackOverflow](https://stackoverflow.com/questions/71070617/spel-with-query-in-spring-data-elasticsearch)

puppylpg:

Closes #2083


src/test/java/org/springframework/data/elasticsearch/repositories/custommethod/QueryParam.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.springframework.data.elasticsearch.repositories.custommethod;
+
+/**
+ * Used as a parameter referenced by SpEL in query method.
+ *
+ * @param q content
+ * @author Haibo Liu
+ */
+public record QueryParam(String q) {
```
sothawo, line 24:
```suggestion
public record QueryParameter(String value) {
```
No need to shorten the name. I think I'd prefer to have the `q` renamed to `value`.

src/test/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQueryUnitTests.java
```diff
@@ -187,9 +258,21 @@ private interface SampleRepository extends Repository<Person, String> {
 		@Query("{ 'bool' : { 'must' : { 'term' : { 'name' : '?0' } } } }")
 		Person findByName(String name);
 
+		@Query("{ \"bool\" : { \"must\" : { \"term\" : { \"name\" : \"#{#name}\" } } } }")
+		Person findByNameSpEL(String name);
+
+		@Query("{ 'bool' : { 'must' : { 'term' : { 'name' : '#{#param.q}' } } } }")
```
sothawo, line 264: formatting

src/test/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQueryUnitTests.java
```diff
@@ -187,9 +258,21 @@ private interface SampleRepository extends Repository<Person, String> {
...
+		Person findByParameterPropertySpEL(QueryParam param);
+
 		@Query("{ 'bool' : { 'must' : { 'terms' : { 'name' : ?0 } } } }")
 		Person findByNameIn(ArrayList<String> names);
 
+		@Query("{ 'bool' : { 'must' : { 'terms' : { 'name' : #{#names} } } } }")
```
sothawo, line 270: formatting

src/test/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQueryUnitTests.java
```diff
@@ -187,9 +258,21 @@ private interface SampleRepository extends Repository<Person, String> {
...
+		Person findByNamesSpEL(ArrayList<String> names);
+
+		@Query("{ 'bool' : { 'must' : { 'terms' : { 'name' : #{#names.![q]} } } } }")
```
sothawo, line 273: formatting

src/test/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchStringQueryUnitTests.java
```diff
@@ -83,6 +86,73 @@ public void shouldReplaceRepeatedParametersCorrectly() throws Exception {
...
 	}
 
+	@Test
+	public void shouldReplaceParametersSpEL() throws Exception {
+
+		org.springframework.data.elasticsearch.core.query.Query query = createQuery("findByNameSpEL", "Luke");
+
+		assertThat(query).isInstanceOf(StringQuery.class);
+		assertThat(((StringQuery) query).getSource())
+				.isEqualTo("{ \"bool\" : { \"must\" : { \"term\" : { \"name\" : \"Luke\" } } } }");
```
sothawo, line 96: Like I wrote further down at the definition of the method, I'd prefer to have this formatted. We should then probably use `JSONAssert` to verify the equality in these tests.

src/test/java/org/springframework/data/elasticsearch/repositories/custommethod/CustomMethodRepositoryIntegrationTests.java
```diff
@@ -1940,6 +2023,97 @@ public interface SampleCustomMethodRepository extends ElasticsearchRepository<Sa
...
+				      {
+				        "term":{
+				          "type":"#{#entity.q}"
+				        }
+				      }
+				    ]
+				  }
+				}
+				""")
+		SearchHits<SampleEntity> queryByParameterPropertySpEL(QueryParam entity);
```
sothawo, line 2054: Entities are what is returned in the `SearchHits`. I think it would be better to name the parameter `parameter` here.

src/test/java/org/springframework/data/elasticsearch/repositories/custommethod/CustomMethodRepositoryIntegrationTests.java
```diff
@@ -1940,6 +2023,97 @@ public interface SampleCustomMethodRepository extends ElasticsearchRepository<Sa
...
+				      {
+				        "terms":{
+				          "type": #{#entities.![q]}
+				        }
+				      }
+				    ]
+				  }
+				}
+				""")
+		SearchHits<SampleEntity> queryByParameterPropertyCollectionSpEL(Collection<QueryParam> entities);
```
sothawo, line 2099: same as above

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchStringValueToStringConverter.java
```diff
@@ -0,0 +1,52 @@
...
+	@Override
+	public Set<ConvertiblePair> getConvertibleTypes() {
+		return Collections.singleton(new ConvertiblePair(String.class, String.class));
+	}
+
+	@Override
+	@Nullable
+	public Object convert(@Nullable Object source, TypeDescriptor sourceType, TypeDescriptor targetType) {
+		if (source == null) {
+			return "null";
```
sothawo, line 43: why not returning `null`?

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchValueSpELConversionService.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.core.convert.ConversionService;
+import org.springframework.core.convert.converter.ConverterRegistry;
+import org.springframework.core.convert.support.DefaultConversionService;
+import org.springframework.data.util.Lazy;
+
+/**
+ * To supply a {@link ConversionService} with custom converter to handle SpEL value in elasticsearch query.
+ *
+ * @author Haibo Liu
+ */
```
sothawo, line 27:
```suggestion
  @since 5.3
 */
```

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchStringValueToStringConverter.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Collections;
+import java.util.Set;
+import java.util.regex.Matcher;
+
+/**
+ * Values in elasticsearch query may contain quotations and should be escaped when converting.
+ * However, the converter should only be used in this situation, rather than common string to string conversions.
+ *
+ * @author Haibo Liu
+ */
```
sothawo, line 31:
```suggestion
  @since 5.3
 */
```

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchCollectionValueToStringConverter.java
```diff
@@ -0,0 +1,76 @@
...
+ * Convert a collection into string for value part of the elasticsearch query.
+ * <p>
+ * The string should be wrapped with square brackets, with each element quoted therefore
+ * escaped(by {@link ElasticsearchStringValueToStringConverter}) if quotations exist in the original element.
+ * <p>
+ * eg: The value part of an elasticsearch terms query should looks like {@code ["hello \"Stranger\"","Another string"]},
+ * for query {@code { 'bool' : { 'must' : { 'terms' : { 'name' : ["hello \"Stranger\"","Another string"] } } } }}
+ *
+ * @author Haibo Liu
+ */
```
sothawo, line 38:
```suggestion
  @since 5.3
 */
```

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchCollectionValueToStringConverter.java
```diff
@@ -0,0 +1,76 @@
...
+		return Collections.singleton(new ConvertiblePair(Collection.class, String.class));
+	}
+
+	@Override
+	@Nullable
+	public Object convert(@Nullable Object source, TypeDescriptor sourceType, TypeDescriptor targetType) {
+		if (source == null) {
+			return "[]";
+		}
+		Collection<?> sourceCollection = (Collection<?>) source;
```
sothawo, line 60: Need a check here if `source` is a collection at all

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchCollectionValueToStringConverter.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.StringJoiner;
+
+/**
+ * Convert a collection into string for value part of the elasticsearch query.
+ * <p>
+ * The string should be wrapped with square brackets, with each element quoted therefore
```
sothawo, line 31: quoted elements means Strings? What about other types, i.e. numeric ones?

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchCollectionValueToStringConverter.java
```diff
@@ -0,0 +1,76 @@
...
+		Collection<?> sourceCollection = (Collection<?>) source;
+		if (sourceCollection.isEmpty()) {
+			return "[]";
+		}
+		StringJoiner sb = new StringJoiner(DELIMITER, "[", "]");
+		for (Object sourceElement : sourceCollection) {
+			Object targetElement = this.conversionService.convert(
+					sourceElement, sourceType.elementTypeDescriptor(sourceElement), targetType);
+			if (sourceElement instanceof String) {
+				sb.add("\"" + targetElement + "\"");
```
sothawo, line 69: `targetElement` can be `null` at this place; the `+` and the `valueOf` will convert this to `"null"; is that ok or should that produce an error?

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchQueryMethod.java
```diff
@@ -69,9 +69,9 @@
  */
 public class ElasticsearchQueryMethod extends QueryMethod {
 
-	// the following 2 variables exits in the base class, but are private. We need them for
+	// the following 2 variables exists in the base class, but are private. We need them for
```
sothawo, line 72:
```suggestion
	// the following 2 variables exist in the base class, but are private. We need them for
```

src/main/java/org/springframework/data/elasticsearch/repository/query/ElasticsearchQueryMethod.java
```diff
@@ -69,9 +69,9 @@
...
 	// correct handling of return types (SearchHits), so we have our own values here.
-	// Alas this means that we have to copy code that initializes these variables and in the
+	// Also, this means that we have to copy code that initializes these variables and in the
```
sothawo, line 74:
```suggestion
	// This means that we have to copy code that initializes these variables and in the
```
get rid of that word

src/main/antora/modules/ROOT/pages/elasticsearch/repositories/elasticsearch-repository-queries.adoc
```diff
@@ -361,3 +361,174 @@ would make an https://www.elastic.co/guide/en/elasticsearch/reference/current/qu
...
+      }
+    ]
+  }
+}
+----
+====
+
+.accessing entity property.
+====
+Supposing that we have the following entity class as query parameter type:
```
sothawo, line 413: I wrote that on the `QueryParam` class as well: this is not an entity and we should not name it as such. Entities in Spring Data are the objects stored in and retrieved from the store.

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchStringValueToStringConverter.java
puppylpg: Assuming that this is the query:
```json
{
  "query": {
    "term": {
      "nickname": {
        "value": null
      }
    }
  }
}
```
Elasticsearch can't accept `null` as value and will produce:
```json
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "value cannot be null"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "value cannot be null"
  },
  "status" : 400
}
```
So there should be two choices for us:
1. convert `null` to string `"null"` or empty `""`, and this will avoid the error;
2. throw an invalid value error directly;

There is [already a similar way](https://docs.spring.io/spring-data/elasticsearch/reference/elasticsearch/repositories/elasticsearch-repository-queries.html#elasticsearch.query-methods.at-query) to insert query parameters into placeholder(`?0`, ...), and the actually implementation in `StringQueryUtil` just treat `null` string as `"null"`:

https://github.com/spring-projects/spring-data-elasticsearch/blob/8f745b19d14ca8af07cda4161781deda305b873d/src/main/java/org/springframework/data/elasticsearch/repository/support/StringQueryUtil.java#L85

so I obey the same implementation here.

However, I now think this isn't appropriate because a `nickname` value may really be string `"null"` and this'll cause the unexpected results.

**I prefer to raise an error for `null` string now.**

When it comes to collection, things can be different. I think:
1. a `null` collection: can be interpreted into `[]`
2. a collection with `null` value `["a", null]`: we can ignore the `null` value in the collection and just keep the non-null strings to produce a collection value like `["a"]`

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchCollectionValueToStringConverter.java
puppylpg: I think the `ConversionService` has already checked the type with `getConvertibleTypes` method, so the `source` must be a collection here and no need to check again.

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchCollectionValueToStringConverter.java
puppylpg: The non string types will be converted correctly by other converters in `ConversionSerivce`. Take integers as the example:
```json
				{
				  "bool":{
				    "must":{
				      "terms":{
				        "name": [1, 2, 3]
				      }
				    }
				  }
				}
```
I'll add more tests to prove this.

puppylpg: And the comment is also changed to avoid the misleading.

src/main/java/org/springframework/data/elasticsearch/repository/support/ElasticsearchStringValueToStringConverter.java
puppylpg: The default `ConversionSerivce` will handle the conversion for `null` source value, rather than a specific converter, so I put the null check into where the SpEL expression is evaluated.

puppylpg:
Review suggestions are accepted and more tests has been added. Reactive parts will be handle later.

sothawo (approved):

PR #2889, opened 2024-04-09
Support has_child and has_parent queries.

### LINKED ISSUE #1472: Add has_child and has_parent type queries [DATAES-900]
**[sothawo](https://jira.spring.io/secure/ViewProfile.jspa?name=sothawo)** opened **[DATAES-900](https://jira.spring.io/browse/DATAES-900?redirect=false)** and commented

After [DATAES-433](https://jira.spring.io/browse/DATAES-433) introduced the join_type into mapping and object storage, the has_child and has_parent queries need to be introduced

---
No further details from [DATAES-900](https://jira.spring.io/browse/DATAES-900?redirect=false)

youssef3wi:

Closes #1472


src/main/java/org/springframework/data/elasticsearch/client/elc/CriteriaQueryProcessor.java
```diff
@@ -397,4 +432,48 @@ public static String escape(String s) {
 		return sb.toString();
 	}
 
+    /**
+     * Convert a spring-data-elasticsearch {@literal query} to an Elasticsearch {@literal query}.
+     * <p>
+     * The reason it was copied from {@link RequestConverter#getQuery(org.springframework.data.elasticsearch.core.query.Query, Class)} is because of a circular dependency.
```
sothawo, line 438: I don't think we should do that duplication. If both `RequestConverter` and `CriteriaQueryProcessor` need that code, it should be extracted into a class that is used by both.

sothawo (changes requested):

src/test/java/org/springframework/data/elasticsearch/core/ElasticsearchIntegrationTests.java
```diff
@@ -49,6 +49,7 @@
 import org.springframework.data.elasticsearch.annotations.*;
 import org.springframework.data.elasticsearch.annotations.Field;
 import org.springframework.data.elasticsearch.annotations.ScriptedField;
+import org.springframework.data.elasticsearch.client.elc.NativeQuery;
```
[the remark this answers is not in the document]
youssef3wi: Yes, I believe that for the build, it is more specific and I don't think it needs to be moved to the derived class.

src/main/java/org/springframework/data/elasticsearch/client/elc/AbstractQueryProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+            if (nativeQuery.getQuery() != null) {
+                esQuery = nativeQuery.getQuery();
+            } else if (nativeQuery.getSpringDataQuery() != null) {
+                esQuery = getEsQuery(nativeQuery.getSpringDataQuery(), queryConverter);
+            }
+        } else {
+            throw new IllegalArgumentException("unhandled Query implementation " + query.getClass().getName());
+        }
+
+        Assert.notNull(esQuery, "query must not be null.");
```
sothawo, line 69: Did you run the tests? This assertion breaks 62 integration tests; the query at this place can be null

sothawo (changes requested):
One nullability issue, looks good otherwise

sothawo (approved):

PR #2905, opened 2024-05-02
Add support for index aliases.

### LINKED ISSUE #2599: Set alias for an index with annotation upon creation.
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html 

The request body for index creation takes a mapping object, a setting object and an alias object. Today we can define the mapping and setting by annotation upon index creation.

It would be nice to be able to do this for alias as well. 

@Alias annotation to define aliases when creating index.

youssef3wi:

Closes #2599

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Alias.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Alias.java
@@ -29,7 +29,7 @@ import java.lang.annotation.Target;
  * Identifies an alias for the index.
  *
  * @author Youssef Aouichaoui
- * @since 5.3
+ * @since 5.4
  */
 @Inherited
 @Retention(RetentionPolicy.RUNTIME)
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Alias.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Alias.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.annotations;
+
+
+import org.springframework.core.annotation.AliasFor;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Identifies an alias for the index.
+ *
+ * @author Youssef Aouichaoui
+ * @since 5.3
+ */
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+@Repeatable(Aliases.class)
+public @interface Alias {
+    /**
+     * @return Index alias name.
+     */
+    String value();
+
+    /**
+     * @return Query used to limit documents the alias can access.
+     */
+    Query filter() default @Query;
+
+    /**
+     * @return Used to route indexing operations to a specific shard.
+     */
+    String indexRouting() default "";
+
+    /**
+     * @return Used to route indexing and search operations to a specific shard.
+     */
+    String routing() default "";
+
+    /**
+     * @return Used to route search operations to a specific shard.
+     */
+    String searchRouting() default "";
+
+    /**
+     * @return Is the alias hidden?
+     */
+    boolean isHidden() default false;
+
+    /**
+     * @return Is it the 'write index' for the alias?
+     */
+    boolean isWriteIndex() default false;
+}
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Alias.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Alias.java
@@ -39,7 +39,11 @@ public @interface Alias {
     /**
      * @return Index alias name.
      */
+    @AliasFor("alias")
     String value();
+    
+    @AliasFor("value")
+    String alias()
 
     /**
      * @return Query used to limit documents the alias can access.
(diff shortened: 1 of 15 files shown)


src/main/java/org/springframework/data/elasticsearch/annotations/Alias.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+@Repeatable(Aliases.class)
+public @interface Alias {
+    /**
+     * @return Index alias name.
+     */
+    String value();
```
sothawo, line 42:
```suggestion
    @AliasFor("alias")
    String value();
    
    @AliasFor("value")
    String alias()
```
With this the user can choose between `value=""` or `alias=""` we keep the Elasticsearch naming with that

sothawo (changes requested):
Sorry it took me a while to get to this.

I don't think that `AliasCoordinates` should be added to `IndexCoordinates`. `IndexCoordinates` is used throughout the library wherever an index name or names are passed. Adding the alias information might cause the user to think he might add some alias info to other functions where this is definitely not. I think that the alias information should be passed to the `IndicesTemplate.create()` and `ReactiveIndicesTemplate.create()` functions like it's done with mapping and settings. This is something only relevant during index creation.

Also I am not sure about reusing the `@Query` annotation. In Spring Data context this normally only exists on repository methods. The documentation for `@Query` says that the query can contain parameters like `?0`, this of course is not valis in the index creation process. I'd rather suggest to creat a new `@Filter` annotation that is only used in the index creation context.

sothawo (changes requested):
Sorry, have been pretty busy (and sick in between), just a first remark, had no time to look further

src/main/java/org/springframework/data/elasticsearch/core/mapping/AliasCoordinates.java
```diff
@@ -0,0 +1,232 @@
...
+
+import java.util.Objects;
+
+/**
+ * Immutable Value object encapsulating index alias(es).
+ *
+ * @author Youssef Aouichaoui
+ * @since 5.3
+ */
+public class AliasCoordinates {
```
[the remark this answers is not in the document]
youssef3wi: Good wishes for recovery 🙂 
Just to avoid conflicts with `co.elastic.clients.elasticsearch.indices.Alias`.

src/main/java/org/springframework/data/elasticsearch/core/mapping/CreateIndexSettings.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Nullable
+    public Document getMapping() {
+        return mapping;
+    }
+
+    public static class Builder {
+        private IndexCoordinates indexCoordinates;
+        private Set<Alias> aliases = new HashSet<>();
```
sothawo, line 72:
```suggestion
        private final Set<Alias> aliases = new HashSet<>();
```

src/main/java/org/springframework/data/elasticsearch/core/mapping/CreateIndexSettings.java
```diff
@@ -0,0 +1,117 @@
...
+
+        private Map<String, Object> settings;
+
+        @Nullable
+        private Document mapping;
+
+        public Builder withIndexCoordinates(IndexCoordinates indexCoordinates) {
```
sothawo, line 79: the index coordinates are mandatory for the `CreateIndexSettings` class, right? I then would change this to a constructor `Builder(IndexCoordinates indexCoordinates)` so that it's not possible to create a builder without it.

src/main/java/org/springframework/data/elasticsearch/core/mapping/CreateIndexSettings.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private CreateIndexSettings(Builder builder) {
+        this.indexCoordinates = builder.indexCoordinates;
+        this.aliases = builder.aliases;
+
+        this.settings = builder.settings;
+        this.mapping = builder.mapping;
+    }
+
+    public static Builder builder() {
```
sothawo, line 49: see my comment in the `Builder` class, this should take the index coordinates

src/main/java/org/springframework/data/elasticsearch/core/mapping/CreateIndexSettings.java
```diff
@@ -0,0 +1,117 @@
...
+    @Nullable
+    public Document getMapping() {
+        return mapping;
+    }
+
+    public static class Builder {
+        private IndexCoordinates indexCoordinates;
+        private Set<Alias> aliases = new HashSet<>();
+
+        private Map<String, Object> settings;
```
sothawo, line 74:
```suggestion
        @Nullable
        private Map<String, Object> settings;
```

src/main/java/org/springframework/data/elasticsearch/core/mapping/CreateIndexSettings.java
```diff
@@ -0,0 +1,117 @@
...
+ * Encapsulating index mapping fields, settings, and index alias(es).
+ *
+ * @author Youssef Aouichaoui
+ * @since 5.3
+ */
+public class CreateIndexSettings {
+    private final IndexCoordinates indexCoordinates;
+    private final Set<Alias> aliases;
+
+    private final Map<String, Object> settings;
```
sothawo, line 36:
```suggestion
    @Nullable
    private final Map<String, Object> settings;
```

src/main/java/org/springframework/data/elasticsearch/core/mapping/ElasticsearchPersistentEntity.java
```diff
@@ -41,6 +43,7 @@
 public interface ElasticsearchPersistentEntity<T> extends PersistentEntity<T, ElasticsearchPersistentProperty> {
 
 	IndexCoordinates getIndexCoordinates();
+	Set<Alias> getAliases();
```
sothawo, line 46: javadoc and `@since` please

src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java
```diff
@@ -138,6 +142,37 @@ public IndexCoordinates getIndexCoordinates() {
 		return resolve(IndexCoordinates.of(getIndexName()));
 	}
 
+	@Override
+	public Set<Alias> getAliases() {
```
sothawo, line 146: please add an `aliases` property to the class and either intialize it from the constructor or make it ` Lazy`, check for example the `settingsParameter` property. This should only be evaluated once.

sothawo (changes requested):
nice work, thank you. 

Some minor things. And the reactive integration test is missing

sothawo (approved):

PR #2920, opened 2024-05-20
Add knn search parameter and remove knn query

### LINKED ISSUE #2919: Add knn search parameter and remove knn query
`KnnQuery` and `KnnSearch` are two different classes in elasticsearch java client and are for different queries:
- `KnnSearch`: is [the top level `knn` query](https://www.elastic.co/guide/en/elasticsearch/reference/8.13/search-search.html#search-api-knn) in the elasticsearch request;
- `KnnQuery`: is [the `knn` query inside `query` clause](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-knn-query.html);

They both have different parameters and different classes in elasticsearch java client. However, in the current spring data elasticsearch implementation, they are mixed together and are all converted into `KnnSearch` finally.

So I wanna submit a PR to:
1. remove `KnnQuery` in `NativeQuery`: and I think this clause should be constructed inside the `co.elastic.clients.elasticsearch._types.query_dsl.Query` in `NativeQuery` by user manually;
2. add `withKnnSearches` method in `NativeQueryBuilder`: this may be a loss without which `KnnSearch` can't be constructed by user;
3. add knn search mapping parameters in elasticsearch 8;

If there are questions about the issues above, we can discuss further.

puppylpg:

Closes #2919

### DIFF
--- a/src/main/antora/modules/ROOT/pages/migration-guides/migration-guide-5.3-5.4.adoc
+++ b/src/main/antora/modules/ROOT/pages/migration-guides/migration-guide-5.3-5.4.adoc
@@ -6,6 +6,17 @@ This section describes breaking changes from version 5.3.x to 5.4.x and how remo
 [[elasticsearch-migration-guide-5.3-5.4.breaking-changes]]
 == Breaking Changes
 
+[[elasticsearch-migration-guide-5.3-5.4.breaking-changes.knn-search]]
+=== knn search
+The `withKnnQuery` method in `NativeQueryBuilder` has been replaced with `withKnnSearches` to build a `NativeQuery` with knn search.
+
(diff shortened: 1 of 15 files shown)


sothawo (changes requested):
Please change the `@since` values to `@since 5.4`. I had to do this change for 5.3.1 as well because Elasticsearch once again had a breaking change in their API in a patch release. Without changing our 5.3.x branch it would not be possible to upgrade Elasticsearch with security fixes on their 8.13 branch.

But the addition of all these classes and removal of a method from the API is a breaking change which we only will do on the main/5.4 branch. Please document these changes on the mogration guid from 5.3 to 5.4, I just added this document to the main branch, you'll need to rebase.

Besides that, it looks fine to me.

Issue #2942, opened 2024-07-18
[FEATURE PROPOSAL] Allow to customize the mapped type for @InnerField and @Field annotations


reta:
In Spring Data OpenSearch [1], which is built on top of Spring Data Elasticsearch, we are  running into cases where both projects differ in incompatible ways. One of such cases is supported mapping field types (baked by `org.springframework.data.elasticsearch.annotations.FieldType`), for example:

```
Flattened("flattened"), //
```

would have to become 

```
Flat_Object("flat_oblect"), //
```

for OpenSearch. 

One of the simplest (but no necessarily good) ways to support Elasticsearch and OpenSearch is to maintain the union of both (`Flattened` and `Flat_Object`): solves the problem but adds confusion. 

One of (arguably)  better alternatives is to add optional `mappedName` attribute to `@InnerField` and` @Field` annotations (and model classes respectively), to address the possible divergence, for example:

```
@Field(type = FieldType.Flattened, mappedName = "flat_oblect") String flattenedField;
```

It would require a bit more work for OpenSearch users but not the Elasticsearch ones. The change would be non-breaking (defaults to "" and `FieldType::getMappedType`) and work seamlessly. 

@sothawo I am wondering if you (or/and other maintainers) would be open to accept such feature proposal (I will be taking care of all the necessary work if greenlighted). 

Thank you!

[1] https://github.com/opensearch-project/spring-data-opensearch

sothawo:
Hi, sure, provide a PR for this.

reta:
(just for the record, it is just a coincidence but I am off next week, will pick this issue once back)

PR #2950, opened 2024-07-31
Allow to customize the mapped type name for @InnerField and @Field annotations

reta:
In Spring Data OpenSearch [1], which is built on top of Spring Data Elasticsearch, we are  running into cases where both projects differ in incompatible ways. One of such cases is supported mapping field types (baked by `org.springframework.data.elasticsearch.annotations.FieldType`), for example:

```
Flattened("flattened"), //
```

would have to become 

```
Flat_Object("flat_oblect"), //
```

for OpenSearch. 

This pull request adds optional `mappedName` attribute to `@InnerField` and` @Field` annotations (and model classes respectively), to address the possible divergence, for example:

```
@Field(type = FieldType.Flattened, mappedName = "flat_oblect") String flattenedField;
```

It would require a bit more work for OpenSearch users but not the Elasticsearch ones. The change would be non-breaking (defaults to `""` and `FieldType::getMappedType`) and work seamlessly. 

- [X] **There is a ticket in the bug tracker for the project in our [issue tracker](https://github.com/spring-projects/spring-data-elasticsearch/issues)**. Add the issue number to the _Closes #issue-number_ line below

Closes https://github.com/spring-projects/spring-data-elasticsearch/issues/2942

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
@@ -243,9 +243,9 @@ public @interface Field {
 	boolean storeEmptyValue() default true;
 	
 	/**
-	 * overrides the mapping field type which otherwise will be taken from corresponding {@link FieldType}
+	 * overrides the field type in the mapping which otherwise will be taken from corresponding {@link FieldType}
 	 *
 	 * @since 5.4
 	 */
-	String mappedName() default "";
+	String mappedTypeName() default "";
 }
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
@@ -38,6 +38,7 @@ import org.springframework.core.annotation.AliasFor;
  * @author Morgan Lutz
  * @author Sascha Woo
  * @author Haibo Liu
+ * @author Andriy Redko
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.FIELD, ElementType.ANNOTATION_TYPE, ElementType.METHOD })
@@ -240,4 +241,11 @@ public @interface Field {
 	 * @since 5.1
 	 */
 	boolean storeEmptyValue() default true;
+	
+	/**
+	 * overrides the mapping field type which otherwise will be taken from corresponding {@link FieldType}
+	 *
+	 * @since 5.4
+	 */
+	String mappedName() default "";
 }
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
@@ -30,6 +30,7 @@ import java.lang.annotation.Target;
  * @author Brian Kimmig
  * @author Morgan Lutz
  * @author Haibo Liu
+ * @author Andriy Redko
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.ANNOTATION_TYPE)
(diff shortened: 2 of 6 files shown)


src/test/java/org/springframework/data/elasticsearch/core/index/MappingBuilderIntegrationTests.java
```diff
@@ -77,6 +78,12 @@ void cleanup() {
 		operations.indexOps(IndexCoordinates.of(indexNameProvider.getPrefix() + "*")).delete();
 	}
 
+	@Test
+	public void shouldSupportAllTypes() {
```
reta, line 82: This one is not strictly related but `EntityWithAllTypes` was not used in tests at all, fixing that

reta:
@sothawo would really appreciate a review when you have some time, thank you

sothawo:
tomorrow, been busy this week 😁

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -240,4 +241,11 @@
 	 * @since 5.1
 	 */
 	boolean storeEmptyValue() default true;
+	
+	/**
+	 * overrides the mapping field type which otherwise will be taken from corresponding {@link FieldType}
```
sothawo, line 246:
```suggestion
	 * overrides the field type in the mapping which otherwise will be taken from corresponding {@link FieldType}
```
Seems a little clearer for me

src/main/java/org/springframework/data/elasticsearch/annotations/InnerField.java
```diff
@@ -171,4 +172,11 @@
 	 * @since 5.4
 	 */
 	KnnIndexOptions[] knnIndexOptions() default {};
+	
+	/**
+	 * overrides the mapping field type which otherwise will be taken from corresponding {@link FieldType}
```
sothawo, line 177: same as above

src/main/java/org/springframework/data/elasticsearch/core/index/MappingBuilder.java
```diff
@@ -210,7 +212,7 @@ private void writeTypeHintMapping(ObjectNode propertiesNode) throws IOException
 		}
 
 		private void mapEntity(ObjectNode objectNode, @Nullable ElasticsearchPersistentEntity<?> entity,
-				boolean isRootObject, String nestedObjectFieldName, boolean nestedOrObjectField, FieldType fieldType,
+				boolean isRootObject, String nestedObjectFieldName, boolean nestedOrObjectField, FieldType fieldType, String mappedName,
```
sothawo, line 215:
```suggestion
				boolean isRootObject, String nestedObjectFieldName, boolean nestedOrObjectField, FieldType fieldType, String fieldTypemappedName,
```
might be mistaken as as mapped name of the entity

src/main/java/org/springframework/data/elasticsearch/annotations/Field.java
```diff
@@ -240,4 +241,11 @@
...
+	 *
+	 * @since 5.4
+	 */
+	String mappedName() default "";
```
sothawo, line 250:
```suggestion
	String mappedTypeName() default "";
```
to not confuse with the field name

sothawo (changes requested):
Thanks for the PR, I just think that the `mappedName` should be called `mappedTypeName` so that it will not be confused with the field name, that is mapped as well from the Java property name to the `Field.name` attribute

reta:
> Thanks for the PR, I just think that the `mappedName` should be called `mappedTypeName` so that it will not be confused with the field name, that is mapped as well from the Java property name to the `Field.name` attribute

Thanks a lot for the review @sothawo , comments addressed! Thanks again!

PR #2964, opened 2024-08-15
fix: upgrade version to 8.15.0

### LINKED ISSUE #2963: Upgrade to Elasticsearch 8.15.

JunHyungJang:
This is my first contribute to spring-data-elasticsearch

I upgrade the version from 8.14.3 -> 8.15.0 as the issue mentioned

- close #2963

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
 		<springdata.commons>3.4.0-SNAPSHOT</springdata.commons>
 
 		<!-- version of the ElasticsearchClient	-->
-		<elasticsearch-java>8.14.3</elasticsearch-java>
+		<elasticsearch-java>8.15.0</elasticsearch-java>
 
 		<blockhound-junit>1.0.9.RELEASE</blockhound-junit>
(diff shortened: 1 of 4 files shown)


sothawo:
Thanks for the contribution, but updating breaks the `org.springframework.data.elasticsearch.client.elc.RequestConverter.reindex(ReindexRequest reindexRequest,
			boolean waitForCompletion)` method, as the `InlineScript` that is used in this method isn't available anymore in 8.15.0.
This can be either a bug in the Elasticsearch client (didn't have time to read the release notes), or we must see how to implement this functionality in a different way.

But the code doesn't even compile after the dependency update

JunHyungJang:
@sothawo  Thanks for the comment
I will figure out how to fix it and able to compile

sothawo:
Sorry for stepping in and taking this over in my own PR. I had a look at this and found that there were breaking changes in the Elasticsearch API in the script API and in the range API. But there was also a necessary adjustment in the reactive client code that was pretty ugly to find.

PR #2969, opened 2024-08-20
Support for DynamicTemplates.

### LINKED ISSUE #2774: Add annotation similar to @JsonUnwrapped on document fields
This is a feature request.

I'm migrating a platform from Apache Solr to ElasticSearch and I'm struggling in creating dynamic fields like Apache Solr used to have. I need facets (aka "aggregations" in ElasticSearch) on these dynamic fields. On Solr that was easily done with an annotation on a Map:

```java
@Dynamic("*_str")
@Field("attributes")
private Map<String, String> attributes;
```

Instead, with ElasticSearch, I need to create a dynamic template for the index like this:

```json
{
    "mappings": {
        "dynamic_templates": [
            {
                "_str": {
                    "match": "*_str",
                    "match_mapping_type": "string",
                    "mapping": {
                        "type": "keyword"
                    }
                }
            }
        ]
    }
}
```

The main issue is that when I now save a document having

```java
@Field("attributes")
private Map<String, String> attributes;
```

, Spring is inputting a payload like

```json
{
  // other document fields above
  "attributes": {
    "code1_str":"value1",
    "code2_str":"other-value"
  },
  // other document fields below
}
```

and I can't do any aggregation query to get a result like

```json
{
    "aggregations": {
        "code1_str": {
            "doc_count_error_upper_bound": 0,
            "sum_other_doc_count": 0,
            "buckets": [
                {
                    "key": "value1",
                    "doc_count": 1
                },
                {
                    "key": "value2",
                    "doc_count": 1
                }
            ]
        },
        "code2_str": {
            "doc_count_error_upper_bound": 0,
            "sum_other_doc_count": 0,
            "buckets": [
                {
                    "key": "other-value",
                    "doc_count": 2
                }
            ]
        }
    }
}
```

The workaround for this issue is atm a custom Spring `Converter<S, T>` for the entire document: immagine having to convert 20 or 30 fields.. this would take a lot of work just because of 1 field serialization/deserialization issue.
A solution could be found Instead in having something like `@JsonUnwrapped` annotation from Jackson library to remove the **attributes** level from the payload, leading to Spring inputting a payload like:

```json
{
  // other document fields above
  "code1_str":"value1",
  "code2_str":"other-value"
  // other document fields below
}
```

Could this be feasible in an upcoming release?

youssef3wi:

Closes #2774 

> I'm uncertain if this will be merged, so the best approach is to use converters.


sothawo:
Thanks for that PR, I finally had some time to have a look.

Restrictions I see in this current draft:
* This currently only is for properties of type `Map<String, Object>` right? 
* Only `match` from the dynamic mapping is supported, we should add `unmatch` as well
* Supporting `path_match` and `path_unmatch` will get pretty ugly, as the mapping from Java property paths to elasticsearch fields in dot notation is hard, we have that in some places, I think that might not be the effort to have it in here
* One thing that is missing is converting the values. The dynamic mapping could for example define a date format and then we need to do a conversion on writing and reading. Or the user might have a custom conversion for all string type or something like that. We'd need to a apply this to these properties as well.
* the reading part in the `readEntity` method should go into its own function, `readEntity` gets too long.

Besides that this would need some documentation not only in the javadoc, but in the reference documentation as well.

youssef3wi:
Honestly I believe it's not good to support this or even improving this draft

sothawo:
Another thought that I just had: the dynamic templates definition must not necessarily come from a `@DynamicTemplates` annotation on the Java entity. It could have been written to the index mapping from outside ot could be contained in a json that is referred to with the `@Mappings` annotation. So it would be necessary to retrieve the current mappping from the cluster to find out the relevant settings here. This would have to be done once and not on every write/read and stored somehow on the entity - doing stuff like this in the ractive world easily gets nasty.. 

Seems quite some effort to add this; if you don't want to go on with this, just drop it.

youssef3wi:
@sothawo I'm not convinced by this draft at the moment.

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
```diff
@@ -602,6 +603,15 @@ protected <T> SearchDocumentResponse.EntityCreator<T> getEntityCreator(ReadDocum
 	// region Entity callbacks
 	protected <T> T maybeCallbackBeforeConvert(T entity, IndexCoordinates index) {
 
+		// get entity metadata
```
sothawo, line 606: This would be called for every conversion of every entity, so that if the user saves 100 entities we have 100 calls.

sothawo (commented):
The more I think about this the more problems seem to arise. We can define on a property that it is part of a dynamic template. But we do not know in which index this entity might be persisted to or retrieved from. So we'd need to load the dynamic template mapping from the cluster for every index name we encounter.
In order to not do this with every save or load, this information would need to be cached. Even if we have this in a cache this add aditional calls even for applications that never need or use this.
For applications that use for example a date in the index name this would mean that over the time we are building up cache data for multiple index names that contain the same data.

I don't think that this architectural and runtime overhead is justified by the single use case - that has only come up ince in all these years that I am maintained SDE

src/main/java/org/springframework/data/elasticsearch/core/AbstractElasticsearchTemplate.java
youssef3wi: Yes, until I find a better place to cache/load metadata 😄

src/main/java/org/springframework/data/elasticsearch/client/elc/IndicesTemplate.java
```diff
@@ -241,6 +248,45 @@ public Map<String, Object> getMapping() {
 		return responseConverter.indicesGetMapping(getMappingResponse, indexCoordinates);
 	}
 
+	@Override
+	public ClusterMapping getClusterMapping() {
+		GetMappingRequest getMappingRequest = requestConverter.indicesGetMappingRequest(IndexCoordinates.of("*"));
```
sothawo, line 253: the user might not have the permission to read that information from all indices in a cluster

src/main/java/org/springframework/data/elasticsearch/client/elc/IndicesTemplate.java
```diff
@@ -241,6 +248,45 @@ public Map<String, Object> getMapping() {
...
+			}
+
+			if (entity == null) {
+				continue;
+			}
+
+			if (mappingEntry.getMappings().containsKey("dynamic_templates")) {
+				Object dynamicTemplates = mappingEntry.getMappings().get("dynamic_templates");
+				if (dynamicTemplates instanceof List<?> value) {
+					getRequiredPersistentEntity(entity).buildDynamicTemplates(value);
```
sothawo, line 284: in line 269 we have that persistentEntity, so there would be no need to call `getRequiredPersistentEntity`

src/main/java/org/springframework/data/elasticsearch/client/elc/IndicesTemplate.java
```diff
@@ -241,6 +248,45 @@ public Map<String, Object> getMapping() {
...
+	 * <p>
+	 * This method is responsible for retrieving and updating the metadata related to the entities.
+	 */
+	private void refreshEntitiesMapping() {
+		ClusterMapping clusterMapping = getClusterMapping();
+		for (ClusterMapping.ClusterMappingEntry mappingEntry : clusterMapping) {
+			// Get entity
+			Class<?> entity = null;
+			for (ElasticsearchPersistentEntity<?> persistentEntity : this.elasticsearchConverter.getMappingContext().getPersistentEntities()) {
+				if (mappingEntry.getName().equals(persistentEntity.getIndexCoordinates().getIndexName())) {
```
sothawo, line 270: the index name defined on the entity could be different than a simple string, that might be a SpEL expression that needs to be evaluated dynamically

sothawo (commented):
There are too many things that won't work (efficiently):
* the user might not have the access rights to read the mappings of all indices in a cluster
* even if so, this only gets the indices from the cluster at one point in time. While the application is running, additional indices can get created in in the cluster, either from outside by some other tool or by some naming, rolling pattern from Elasticsearch defined by index templates. Or the user can have SpEL in the indexname like I described in [this post](https://www.sothawo.com/2020/07/how-to-provide-a-dynamic-index-name-in-spring-data-elasticsearch-using-spel/)

Consider the following case:
The indexname is defined with something like `@Document(indexName = "log-#{T(java.time.LocalDate).now().toString()}")` combined with in index template to define the mapping. This will result in indices named _log-2024-09-4_, _log-2024-9-5_ etc be defined all while the application is running. We would constantly need to reload and adjust the collection of index mapping information.

When searching the user might search the index _log-*_, we'd need to check every returned search document to find from which index it came and would need to then eventually load the index mapping information.

I don't think that we should add this runtime, network traffic and memory overhead for this single mapping usecase

src/main/java/org/springframework/data/elasticsearch/client/elc/IndicesTemplate.java
youssef3wi: Names in **IndexCoordinates** have been resolved

https://github.com/spring-projects/spring-data-elasticsearch/blob/d06c122fd55e55b8886b1d3a3389516b08066784/src/main/java/org/springframework/data/elasticsearch/core/mapping/SimpleElasticsearchPersistentEntity.java#L347

youssef3wi:
I don't see any issues here.
This draft can resolve the problem, but it remains a draft due to the types of conversions.

Are you suggesting closing this draft?

sothawo:
I still see the issues already mentioned:

```
public ClusterMapping getClusterMapping() {
		GetMappingRequest getMappingRequest = requestConverter.indicesGetMappingRequest(IndexCoordinates.of("*"));
``` 
This call might fail as the user might not have access to all indices in a cluster, for example when the cluster serves multiple users/tenants, when kibana is installed which stores data in it's own indices etc. So we cannot guarantee to have the information about an index that we need.

We need to have the index mapping information when converting an entity to a document. But it can be that the index where the document is to be stored does not exist yet. Assume we have an index template for index names like _log-*_ which contains a mapping with a dynamic template, and the user saves some data to _log-2024-09-10_ and this index does not exist yet. We cannot do the mapping resepcting the dynamic template because Elasticsearch will only create that index when the document is sent to the cluster, but that's to0 late for our conversion.

We do not necessarily have the information about an index when reading response data. Let's assume we start the application and read all the index information that we can and that does not fail. To stick with the example we get the index mappings for all the _log-*_ indices until the current date like _log-2024-09-10_. On the next day, the user searches for some data in _log-*_. Now it can be that some log ingestion mechanism has inserted something into _log-2024-09-11_ before the search started. This index now exists, but our application does not know about this. We then could get documents from _log-2024-09-10_ that we can convert and some from _log-2024-09-11_ where we don't have the index mapping in the cache.

As long as these issues are not resolved we cannot reliably offer this as a solution

Issue #3136, opened 2025-07-21
UpdateQuery primaryTerm/seqNo should be long?


TheHett:
`SeqNoPrimaryTerm` provides long datatypes

https://github.com/spring-projects/spring-data-elasticsearch/blob/55d470fc91151d6a927a41fb6dc293f0ef0e1810/src/main/java/org/springframework/data/elasticsearch/core/query/SeqNoPrimaryTerm.java#L43

But `UpdateQuery` accepts Int.  

https://github.com/spring-projects/spring-data-elasticsearch/blob/55d470fc91151d6a927a41fb6dc293f0ef0e1810/src/main/java/org/springframework/data/elasticsearch/core/query/UpdateQuery.java#L45-L46

It may be casted to a certain point, but then it becomes negative.

sothawo:
you're right, in the Elasticsearch Client these are of Type long as well.

PR #3137, opened 2025-07-22
Fixed PrimaryTerm/SeqNo datatypes in UpdateQuery

TheHett:
Fixed incorrect datatype of `primaryTerm/seqNo` fields in UpdateQuery (just changed Int -> Long)

I did not find similar cases in other classes.

Closes https://github.com/spring-projects/spring-data-elasticsearch/issues/3136


sothawo:
Thank you for the PR (and finding that inconsistency).

There are two small additional changes - which I will address in a follow up commit):
- the code in the `RequestConverter` where these fields are used can be cleaned up.
- as this is a breaking API change for users, the documentation needs to be adapted

TheHett:
@sothawo Thanks!

PR #3281, opened 2026-04-24
Refactor routing and null handling with a new getRouting method

### LINKED ISSUE #3278: `routing=null` in queries when using elasticsearch client 9.3
I don't have any special routing for my indices, but I've started having problems with Queries after upgrading to elasticsearch 9.3, as simple queries using findByID would return no results, when the record was definitively there.

On turning on Trace I could see that the query being sent to elasticsearch was 

`GET /dev_directory/_doc/Yy6LtJ0BCdAQ_iYP9L1l?routing=null`

which returned no documents, but

`GET /dev_directory/_doc/Yy6LtJ0BCdAQ_iYP9L1l`

returns as expected.

I'm not sure when this routing field was added, and perhaps it's always been there, but if it is null I don't think it should be appended, as it is now telling elasticsearch to use the string `null` to calculate the shard where the document is stored instead of the ID. As some of our indices have more than 1 shard it means it that chances of hitting the correct shard is `1/n` where `n` is number of shards.

For reference we are using Boot 4.1.0-M4, and data-elasticsearch 6.1.0-M2 In our Dev Environment with elasticsearch 9.3.
It's possible this also impacted 9.2 as well, but we've only just noticed it.

ref : https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-routing-field

Stexxen:
Refactor routing and null handling with a new getRouting method

- [* ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [*] **There is a ticket in the bug tracker for the project in our [issue tracker](https://github.com/spring-projects/spring-data-elasticsearch/issues)**. Add the issue number to the _Closes #issue-number_ line below
- [*] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [*] You submit test cases (unit or integration tests) that back your changes.
- [*] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

Closes #3278

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
+++ b/src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
@@ -74,6 +74,7 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.Optional;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -113,6 +114,7 @@ import org.springframework.util.StringUtils;
  * @author cdalxndr
  * @author scoobyzhang
  * @author Haibo Liu
+ * @author Steven Pearce
  * @since 4.4
  */
 class RequestConverter extends AbstractQueryProcessor {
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
```diff
@@ -1993,6 +1975,10 @@ private String getPersistentEntityId(Object entity) {
 		return null;
 	}
 
+	Optional<String> getRouting(@Nullable String ...routingList) {
+		return Arrays.stream(routingList).filter(StringUtils::hasText).findFirst();
```
sothawo: `Arrays.stream()` does not work with a `null` parameter, this ends in a `Objects.requireNotNull` which throws a `NullPointerException` when calling `getRouting(null)`

sothawo (changes requested):

src/main/java/org/springframework/data/elasticsearch/client/elc/RequestConverter.java
Stexxen: I've swapped out the varargs method and replaced it with 2 methods with 1 and 2 args as all uses only have 1 or 2 possible variables.
I've also a added another test for the single null.

PR #3309, opened 2026-07-17
Allow to provide non-standard (custom) index options for @Field / @InnerField

### LINKED ISSUE #3299: Allow to provide non-standard (custom) index options for @Field / @InnerField
### Problem

The `@Field` / `@InnerField` annotations have very rich but predefined set of attributes (properties). It does work for the majority of the applications but often, especially when Spring Data OpenSearch is involved, it is not enough. The good example is K-NN indexing support, were the algorithms / engines / parameters are vastly different.

### Suggested Solution(s)

The functionality delivered by https://github.com/spring-projects/spring-data-elasticsearch/issues/3254 allows to transparently transform Elasticsearch specific option to OpenSearch ones, and it works really well. The last gap that is left unaddressed is passing the option that only Elasticsearch  or only OpenSearch has. 

There are few options worth considering:
 - add custom annotation (fe `KnnField`, etc), this will require a large amount of changes since Spring Data Elasticsearch only recognizes  `@Field` / `@InnerField` presence
 - add `customIndexOptions` attribute to `@Field` / `@InnerField`
   ```java
    /**  The index option mapper that will be used to populate this custom index option, follows MappingParameters naming */
    public interface IndexOptionMapper {
        void writeIndexOptionTo(CustomIndexOption indexOption, ObjectNode objectNode);
    }

    public @interface CustomIndexOption {
        /**
         * The name of the custom index option 
         */
        String name();
    
        /**
         * The value(s) of the custom index option 
         */
        String[] values() default {};
    
        /**
         * Should the index property be overridden if already present or not 
         */
        boolean overrideIfPresent() default false;

        /**
         * The index option mapper that will be used to populate this custom index option
         */
        Class<? extends IndexOptionMapper> mapper();
    }
    
    public @interface Field {
        ...
        CustomIndexOption[] customIndexOptions() default {};
    }
    
    public @interface InnerField {
        ...
        CustomIndexOption[] customIndexOptions() default {};
    }
   ```
   which `MappingParameters`, if present, would include into final `ObjectNode` as-is (per `IndexOptionMapper` implementation(). 

The sample pull request to showcase the limitations with Elasticsearch <-> OpenSearch conversions: https://github.com/opensearch-project/spring-data-opensearch/pull/626

Usage example:
```java
@Field(type = FieldType.Dense_Vector, dims = 42, mappedTypeName = "knn_vector",
    knnSimilarity = KnnSimilarity.COSINE, 
    customIndexOptions = {
        @CustomIndexOption(name = "method", values = "faiss", overrideIfPresent = true, mapper = StringMapper.class)
    }
) private double @Nullable [] denseVector;
```

@sothawo as always, would love to hear your opinion, if there is no objections from your side, I would send the pull request over, thank you.

reta:
The @Field / @InnerField annotations have very rich but predefined set of attributes (properties). It does work for the majority of the applications but often, especially when Spring Data OpenSearch is involved, it is not enough. The good example is K-NN indexing support, were the algorithms / engines / parameters are vastly different.

Closes #3299

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/CustomIndexOption.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/CustomIndexOption.java
@@ -24,7 +24,7 @@ import org.springframework.data.elasticsearch.core.index.IndexOptionMapper;
  * 
  * @author Andriy Redko
  * 
- * @since 6.1.1
+ * @since 6.2
  */
 public @interface CustomIndexOption {
     /**
--- a/src/main/java/org/springframework/data/elasticsearch/annotations/CustomIndexOption.java
+++ b/src/main/java/org/springframework/data/elasticsearch/annotations/CustomIndexOption.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.elasticsearch.annotations;
+
+import org.springframework.data.elasticsearch.core.index.IndexOptionMapper;
+
+/**
+ * Represents the custom index option, either not provided by the core (fe, custom plugin)
+ * or deviates across different engines.
+ * 
+ * @author Andriy Redko
+ * 
+ * @since 6.1.1
+ */
+public @interface CustomIndexOption {
+    /**
+     * The name of the custom index option 
+     */
+    String name();
+
+    /**
+     * The value(s) of the custom index option 
+     */
+    String[] values() default {};
+
+    /**
+     * Should the index property be overridden if already present or not 
+     */
+    boolean overrideIfPresent() default false;
+    
+    /**
+     * The index option mapper that will be used to populate this custom index option
+     */
+    Class<? extends IndexOptionMapper> mapper();
+}
--- a/src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
+++ b/src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
@@ -1,0 +1,306 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.elasticsearch.core.index;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Arrays;
+
+import org.springframework.data.elasticsearch.annotations.CustomIndexOption;
+
+import tools.jackson.databind.JsonNode;
+import tools.jackson.databind.node.ArrayNode;
+import tools.jackson.databind.node.BigIntegerNode;
+import tools.jackson.databind.node.BooleanNode;
+import tools.jackson.databind.node.ObjectNode;
+import tools.jackson.databind.node.StringNode;
+import tools.jackson.databind.node.DecimalNode;
+
+/**
+ * The collection of predefined {@link IndexOptionMapper}s
+ * 
+ * @author Andriy Redko
+ * 
+ * @since 6.1.1
+ */
+public final class IndexOptionMappers {
+	private IndexOptionMappers() {
+	}
+
+	/**
+	 * Writes the {@link CustomIndexOption} instance into mapping as JSON string (or array of strings) property
+	 */
+	public static final class StringMapper implements IndexOptionMapper {
+	@Override
+		public void writeIndexOptionTo(CustomIndexOption indexOption, ObjectNode objectNode) {
+			final String[] values = indexOption.values();
+			if (values.length == 1) {
+				writePropertyAsString(indexOption.name(), indexOption.overrideIfPresent(), values[0], objectNode);
+			} else if (values.length > 1) {
+				writePropertyAsArray(indexOption.name(), indexOption.overrideIfPresent(), values, objectNode);
+			} else {
+				removeProperty(indexOption.name(), objectNode);
+			}
+		}
+	}
+
+	/**
+	 * Writes the {@link CustomIndexOption} instance into mapping as JSON number (or array of numbers) property
+	 */
+	public static final class NumberMapper implements IndexOptionMapper {
+		@Override
+		public void writeIndexOptionTo(CustomIndexOption indexOption, ObjectNode objectNode) {
+			final Number[] values = toNumbers(indexOption.values());
+			if (values.length == 1) {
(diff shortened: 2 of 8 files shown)


sothawo:
will look at it tomorrow

src/main/java/org/springframework/data/elasticsearch/annotations/CustomIndexOption.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.springframework.data.elasticsearch.core.index.IndexOptionMapper;
+
+/**
+ * Represents the custom index option, either not provided by the core (fe, custom plugin)
+ * or deviates across different engines.
+ * 
+ * @author Andriy Redko
+ * 
+ * @since 6.1.1
```
sothawo: for all the changes: This is an API change and new feature, no bugfix, so it would go into 6.2

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
```diff
@@ -0,0 +1,306 @@
...
+	/**
+	 * Convert the array of strings to array of numbers, throwing {@link NumberFormatException} if the conversion 
+	 * is not possible.
+	 * 
+	 * @param values array of strings
+	 * @return array of numbers
+	 * 
+	 * @throws NumberFormatException
+	 */
+	private static Number[] toNumbers(String[] values) {
```
sothawo: this defines the input and output as non-null arrays of non-null elements. Using jspecify the correct definition would be
```
private static @Nullable Number[] toNumbers(@Nullable String[] values)
```

Can the elements of the input array be `null`? Or is this just an empty array when no values are specified?

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
```diff
@@ -0,0 +1,306 @@
...
+		final Number[] numbers = new Number[values.length];
+		for (int j = 0; j < values.length; ++j) {
+			if (values[j] == null) {
+				numbers[j] = null;
```
sothawo: this is unreachable if the array elements are non-nullable

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
```diff
@@ -0,0 +1,306 @@
...
+					}
+
+					if (Character.digit(s.charAt(i), 10) < 0) {
+						isInteger = false;
+						break;
+					}
+				}
+
+				if (isInteger) {
+					numbers[j] = new BigInteger(s);
```
sothawo: unreachable code, `isInteger` is never set to `true`

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
```diff
@@ -0,0 +1,306 @@
...
+				} else {
+					numbers[j] = new BigDecimal(s);
+				}
+			}
+		}
+
+		return numbers;
+	}
+
```
sothawo, line 135: Instead of doing this parsing and checking, why not:
```
private static @Nullable Number[] toNumbers(@Nullable String[] values) {
    final Number[] numbers = new Number[values.length];
    for (int j = 0; j < values.length; ++j) {
        Number number = null;
        var value = values[j];

        if (value != null) {
            try {
                number = new BigInteger(value);
            } catch (NumberFormatException e) {
                number = new BigDecimal(value);
            }
        }

        numbers[j] = number;
    }
    return numbers;
}
```

and a test that checks the conversion of different strings.

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
```diff
@@ -0,0 +1,306 @@
...
+		}
+	}
+
+	/**
+	 * Writes the {@link CustomIndexOption} instance into mapping as JSON number (or array of numbers) property
+	 */
+	public static final class NumberMapper implements IndexOptionMapper {
+		@Override
+		public void writeIndexOptionTo(CustomIndexOption indexOption, ObjectNode objectNode) {
+			final Number[] values = toNumbers(indexOption.values());
```
sothawo:
```suggestion
			final @Nullable Number[] values = toNumbers(indexOption.values());
```

See my comment in the `toNumbers` method. So here the Number array can contain `null` values

sothawo (commented):
Only some nullability issues/questions in the mapper and info about the affected version

src/main/java/org/springframework/data/elasticsearch/annotations/CustomIndexOption.java
reta: Got it, had my fingers crossed for 6.1.x, but totally understand

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
reta: > Can the elements of the input array be `null`? Or is this just an empty array when no values are specified?

Yes, they can, I don't have a case at hand where it could be useful, but supporting it just in case.

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
reta: Right. the could be, added `@Nullable` as recommended

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
reta: Ah, thank you, my bad, fixed

src/main/java/org/springframework/data/elasticsearch/core/index/IndexOptionMappers.java
reta: So I was debating myself between exception driven flow and the one I implemented (iteration + checks), in general, the exceptions variant is shorter but to be fair, it is somewhat "expensive" way vs just simple checks. I will certainly work, if you have preference towards this option, no objection, I could change it, thanks!

sothawo: Yes, I was thinking about the performance of that first try catch in comparison to self-checking. But this code is not executed often, only when an index mapping is built on creating an index. So I'd prefer to have the shorter version with Exceptions and leave the converting logic to the `Biginteger` and `BigDecimal` classes (recently in my daytime job I had to deal with eastern arabic number literals, and there checking for 0..9 would fail - although as long as we don't pass in Locales we couldn't handle these anyway)

src/main/java/org/springframework/data/elasticsearch/annotations/CustomIndexOption.java
sothawo, line 38: I just tried it, I cannot pass a `null` value. So it might be an empty array, but it does not contain `null`

Image

That means you can remove the null checks in the mappers

reta: I think it is a bit trickier, the annotations could be created programmatically, fe:

```
CustomIndexOption ann = new CustomIndexOption() {
            @Override
            public Class<? extends Annotation> annotationType() {
                return CustomIndexOption.class;
            }
            
            @Override
            public String[] values() {
                return null;
            }
            
            @Override
            public boolean overrideIfPresent() {
                return false;
            }
            
            @Override
            public String name() {
                return null;
            }
            
            @Override
            public Class<? extends IndexOptionMapper> mapper() {
                return null;
            }
        };
```

We don't use such pattern in Spring Data Elasticsearch / OpenSearch, but it technically possible to encounter one.

sothawo: that would make the returned array `null`, not it's elements.

And IntelliJ marks this for me 
image

`CustomIndexOption` specifies it as an non-null array with non-null elements. 

Thinking about it, Id then rather would check that `value` is not null and throw an Exception if one of its values is `null`

reta: Added explicit `requireNotNull` checks, thanks @sothawo 

> that would make the returned array null, not it's elements.

Sorry, I should have been more explicit with example:

```java
            @Override
            public String[] values() {
                return null;
            }
```

or 

```java
            @Override
            public String[] values() {
                return new String[] { "a", null, "b" };
            }
```

both are possible, and should be covered now, thank you

PR #3313, opened 2026-07-24
Spring Cache using Elasticsearch

senocak:
A full Spring Cache using Elasticsearch that implement Spring's cache abstraction on top of Elasticsearch, storing each entry as a document via the store-independent ElasticsearchOperations API.

- [ ] **There is a ticket in the bug tracker for the project in our [issue tracker](https://github.com/spring-projects/spring-data-elasticsearch/issues)**. Add the issue number to the _Closes #issue-number_ line below

Closes #issue-number

### DIFF
--- a/src/main/java/org/springframework/data/elasticsearch/cache/CacheDocument.java
+++ b/src/main/java/org/springframework/data/elasticsearch/cache/CacheDocument.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.elasticsearch.cache;
+
+import org.jspecify.annotations.Nullable;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.elasticsearch.annotations.Document;
+import org.springframework.data.elasticsearch.annotations.Field;
+import org.springframework.data.elasticsearch.annotations.FieldType;
+import org.springframework.data.elasticsearch.annotations.WriteTypeHint;
+
+/**
+ * Internal representation of a persisted cache entry.
+ *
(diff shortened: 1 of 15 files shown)


senocak:
@sothawo Just a friendly follow-up on this PR whenever you have a chance. I'd really appreciate a review, and I'm happy to make any changes if needed. Thanks!

sothawo:
Just a friendly reminder, that it's weekend and the time have to work on Spring Data Elasticsearch is limited by me having a family and hobbies, and during the week I have a normal day time job.

senocak:
@sothawo You're absolutely right, family always comes first. I apologize if my message came across as rushing the review which it wasn't my intention. I was mainly trying to understand whether the feature seems like a worthwhile addition or if it's likely not the right direction for the project. I'm happy to wait until you have time to review it.
Thank you for maintaining the project

sothawo:
I asked the maintainers of other Spring Data modules if they already have a Spring Cache implementation, I'll leave this answer here - at the moment without further comments, I'm on a different issue atm, but will come to this later:

> We do have a cache implementation in Spring Data Redis (RedisCacheManager, see https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/cache/RedisCacheManager.java), and one thing we learned there is that users care a lot about how keys and cache names are composed. Key patterns end up being one of the most frequently configured aspects. Against that background, a fixed @Document(indexName = ElasticsearchCacheManager.DEFAULT_INDEX_NAME) goes in the opposite direction: it settles naming at compile time, which is precisely where we found flexibility to be needed.
>
>The other point is concurrency: For a cache, the relevant scope is not the Cache instance within a single JVM but coordination across JVM boundaries. Emulating eviction or expiry outside of what the store does natively tends to be fragile: if an application is not running, or a connection is temporarily unavailable, the emulated bookkeeping drifts and the cache is left in a stale or inconsistent state.
>
>So my suggestion is to first take a careful look at what Elasticsearch itself provides in terms of functionality suitable for caching (expiry, eviction, and so forth), and then keep the implementation strictly within those boundaries. A narrow implementation that maps directly onto native store semantics is easier to reason about, and we can always widen it later if concrete needs come up. Also, sometimes the conclusion is to not build a feature creates more value than a feature that causes more trouble than good.

senocak:
@sothawo Thanks for taking the time to discuss this with the other Spring Data maintainers and for the detailed feedback. Those are very good points, especially around configurable cache naming and relying on Elasticsearch's native semantics instead of trying to emulate cache behavior at the framework level. My initial implementation was primarily intended as a poc to explore how Spring Cache could be backed by Elasticsearch.

I’m happy to wait until you have time to take a deeper look at the pr. I appreciate the feedback so far, and I'd be grateful for your guidance on how you think the implementation should evolve based on these considerations.

Thank you

